PR #11, opened 2021-07-02
established mongo connection

tamarayustian:


server/api/models/member.js
```diff
@@ -0,0 +1,49 @@
+module.exports = {
```
albertparedandan, line 1: i think this models folder will definitely grow in size as we continue developing. Can you look into putting these models in subfolders? e.g server/api/models/user/member.js

albertparedandan: is everyone in our church considered a member? if so i think it makes more sense to call them user? or something more generic?

if not everyone in our church considered a member, should probably generalise the "user" into something more scalable. 

open to discussing more about this

server/api/models/member.js
```diff
@@ -0,0 +1,49 @@
...
+  attributes: {
+    id: {
+      type: 'string',
```
albertparedandan, line 4: is there a reason why id type is string instead of number?

albertparedandan: actually, on top of that, do we need an id column? isn't it already autogenerated by sails in server/config/models.js?

server/api/models/member.js
```diff
@@ -0,0 +1,49 @@
...
+      columnName: '_id'
+    },
+    createdAt: {
+      type: 'number',
+      autoCreatedAt: true,
+    },
+    updatedAt: {
+      type: 'number',
```
albertparedandan, line 12: probably don't these 2 columns above to be explicitly defined since already defined in server/config/models.js

server/api/models/member.js
```diff
@@ -0,0 +1,49 @@
...
+      type: 'string',
+      required: true
+    },
+    email: {
+      type: 'string',
+      required: true,
+      unique: true,
+      isEmail: true
+    },
+    campus: {
```
albertparedandan, line 25: if this is required, what is campus for someone in Focus lifestage going to be?

server/api/models/member.js
```diff
@@ -0,0 +1,49 @@
...
+      type: 'string',
+      required: true
+    },
+    lifestage: {
+      type: 'string',
+      required: true
+    },
+    isMember: {
+      type: 'ref',
```
albertparedandan, line 34: why is this type ref instead of boolean?

albertparedandan: any reason why this is not required?

server/api/models/member.js
```diff
@@ -0,0 +1,49 @@
...
+      defaultsTo: false
+    },
+    isBaptised: {
```
albertparedandan, line 37: any reason why this is not required?

server/api/models/member.js
```diff
@@ -0,0 +1,49 @@
...
+      type: 'boolean',
+      defaultsTo: false
+    },
+    phoneNumber: {
+      type: 'number',
+      required: true
+    },
+    financeID: {
+      type: 'number'
```
albertparedandan, line 46: correct me if i'm wrong, but based on my understanding from what Joanna explained, a member A can have multiple id's in tithely giving database. If so, i think this shouldn't be a number type

server/api/controllers/member.js/add-member.js
```diff
@@ -0,0 +1,24 @@
...
+          lifestage: 'string',
+          isMember: 'boolean',
+          isBaptised: 'boolean',
+          phoneNumber: 'number',
+          //classesTaken: 'json'
+      }]
+    }
+  },
+  exits: {},
+  fn: async function(inputs, exits) {
```
albertparedandan, line 20: can use `{}` instead to destructure objects. e.g: `async function({ member }, exits) {`

server/api/controllers/member.js/add-member.js
```diff
@@ -0,0 +1,24 @@
+module.exports = {
```
albertparedandan, line 1: similar to the other comment i made in member.js model. Should we consider a more generic name instead of member? like user? also folder name should be member instead of member.js

server/api/controllers/member.js/add-member.js
```diff
@@ -0,0 +1,24 @@
...
+  friendlyName: 'Add members',
+  description: '',
+  inputs: {
+    member: {
```
albertparedandan, line 5: are you trying to create a controller that receives a member type object and saves it? if so, can just make this a json type

```
inputs: {
  member: {
    type: 'json',
    required: true,
    description: 'member object'
  }
}
```

albertparedandan (changes requested):

ksuhartono97 (approved):
albert's comments look good. can merge after those changes go thru

server/api/controllers/member.js/add-member.js
tamarayustian: ohh yea this is what im trying to do. this is helpful

server/api/models/member.js
tamarayustian: oh i didnt know its autogenerated according to that. ok will remove it then

tamarayustian: id type is string because ObjectId generated by mongodb is string in sails

server/api/models/member.js
tamarayustian: i think i was trying to make it tied to the membership table, but after thinking about it more, I'll just have it boolean and false as default
it's not required because I'm not sure if it'll be a required field up in the sign up process for frontend, so i will just make it false for now

server/api/models/member.js
tamarayustian: same as above

server/api/controllers/member.js/add-member.js
tamarayustian: yeah will change to user actually

server/api/models/member.js
tamarayustian: true, it should be more general. updated the schema according to this suggestion and renamed 'member' to 'user'

server/api/models/member.js
tamarayustian: thinking that focus lifestage is free to fill in their previous campus too, or people who have moved on to focus can still have that history of being part of campus. but I'm open to combining campus and lifestage or making these fields not required and they can be empty strings

server/api/models/member.js
tamarayustian: will change it to array of strings then

server/api/models/user.js
```diff
@@ -30,8 +27,11 @@ module.exports = {
       type: 'string',
       required: true
     },
+    LIFEgroup: {
```
albertparedandan, line 30: lifeGroup?

server/api/models/user.js
```diff
@@ -42,8 +42,20 @@ module.exports = {
       type: 'number',
       required: true
     },
+    classesTaken: {
+      type: 'json'
+    },
+    eventsJoined: {
+      type: 'json'
+    },
+    membershipID: {
```
albertparedandan, line 51: membershipId.. same as below

server/api/models/user.js
```diff
@@ -1,16 +1,13 @@
 module.exports = {
```
albertparedandan, line 1: is this also supposed to have email/campus/lifestage/phoneNumber?

albertparedandan (changes requested):

albertparedandan (approved):
LGTM!

PR #17, opened 2021-07-10
8 feature login token management

tyleung:
Basic signup and login functionality works. Need to hook it up with the model/db once that is merged in.
TODO: email verification

### DIFF
--- a/server/api/controllers/auth/post-signup.js
+++ b/server/api/controllers/auth/post-signup.js
@@ -1,0 +1,115 @@
+module.exports = {
+  friendlyName: "Signup",
+
+  description: "Sign up for a new user account.",
+
+  extendedDescription: `This creates a new user record in the database, signs in the requesting user agent
+by modifying its [session](https://sailsjs.com/documentation/concepts/sessions), and
+(if emailing with Mailgun is enabled) sends an account verification email.
+
+If a verification email is sent, the new user's account is put in an "unconfirmed" state
+until they confirm they are using a legitimate email address (by clicking the link in
+the account verification message.)`,
+
+  inputs: {
+    emailAddress: {
+      required: true,
+      type: "string",
+      isEmail: true,
+      description: "The email address for the new account, e.g. m@example.com.",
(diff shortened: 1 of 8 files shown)


server/config/routes.js
```diff
@@ -9,31 +9,30 @@
...
+   *                                                                          *
+   ***************************************************************************/
 
-  'get /api/sermons/get-all-sermons' : 'sermons/get-all-sermons',
-  'get /api/sermons/get-sermon-by-id' : 'sermons/get-sermon-by-id',
-  'get /api/sermons/get-speaker' : 'sermons/get-speaker',
-  'get /api/sermons/get-sermon-series' : 'sermons/get-sermon-series',
+  "post /api/auth/login": "auth/post-login",
+  "post /api/auth/signup": "auth/post-signup",
+  "get /api/sermons/get-all-sermons": "sermons/get-all-sermons",
```
albertparedandan, line 23: can we have new line here just for organising

albertparedandan: can also do it in next PR if you don't want to do it here

server/api/controllers/auth/post-signup.js
```diff
@@ -0,0 +1,115 @@
...
+      description: "The unencrypted password to use for the new account.",
+    },
+
+    fullName: {
+      required: true,
+      type: "string",
+      example: "Frida Kahlo de Rivera",
+      description: "The user's full name.",
+    },
+  },
```
albertparedandan, line 37: think there are other required fields based on the db that we need here

server/api/controllers/auth/post-signup.js
```diff
@@ -0,0 +1,115 @@
...
+
+  exits: {
+    success: {
+      description: "New user account was created successfully.",
+    },
+
+    invalid: {
```
albertparedandan, line 44: should also probably handle other required fields being invalid

albertparedandan (changes requested):

albertparedandan (approved):

PR #19, opened 2021-07-11
create get announcements

Midnite1212:
Finished get all announcements, by id, custom host and routes.


server/api/controllers/sermons/get-all-announcements.js
```diff
@@ -0,0 +1,48 @@
+const { DateTime } = require('luxon');
```
albertparedandan, line 1: small note, PR should merge to release not main branch. I changed yours already so it's fine

server/api/controllers/sermons/get-all-announcements.js
```diff
@@ -0,0 +1,48 @@
...
+const he = require('he');
+
+const transformAnnouncement = (announce) => {
```
albertparedandan, line 4: it'd be good to look into some of the ways we're using caching and helpers to do a lot of this. can talk to you more about it thru call and help walk through it with you

server/api/controllers/sermons/get-all-announcements.js
```diff
@@ -0,0 +1,48 @@
...
+  return {
+    id: announce.id,
+    slug: announce.slug,
+    title : he.decode(announce.title.rendered),
+    startDate: DateTime.fromFormat(announce.date, "LLLL MM dd yyyy"),
```
albertparedandan, line 9: any reason why the formats for the time are different? ideally it should just be `2021-07-04T21:24:03.000+08:00` as an example

server/api/controllers/sermons/get-all-announcements.js
albertparedandan: also maybe `annoucement` instead of `announce`

albertparedandan (changes requested):

server/api/controllers/sermons/get-all-announcements.js
Midnite1212: okk thanks

server/api/controllers/sermons/get-all-announcements.js
Midnite1212: that's the one im not rly sure, cuz the json file format was in unicode like "Fri \u00b7 7\/30\/2021" not sure whats the format for this...

tyleung (dismissed):

tyleung (commented):
use https in custom.js

tyleung (approved):

albertparedandan (approved):

PR #22, opened 2021-07-12
21 feature user api

tamarayustian:
- added nationality field
- fixed user API, including auth
- created view-user


server/api/controllers/auth/post-signup.js
```diff
@@ -35,6 +35,13 @@ the account verification message.)`,
       description: "The user's full name.",
     },
 
+    nationality: {
+      required: true,
+      type: "string",
+      example: "Hong Kong",
```
albertparedandan, line 41: wondering what the datas would look like in the future for this field. "hong_kong"? "hongkong"? will this be a drop down/text field? if drop down do we have a table of countries? not a big deal right now but just came across my mind as i read through this

server/api/controllers/users/update-user.js
```diff
@@ -0,0 +1,24 @@
+module.exports = {
+  friendlyName: 'View users',
```
albertparedandan, line 2: "Update user"

server/api/controllers/users/update-user.js
```diff
@@ -0,0 +1,24 @@
...
+  inputs: {
+    params: {
+      required: false,
+      type: 'json'
+    }
+  },
+
+  exits: {
+    success: {
+      description: "User account returned successfully",
```
albertparedandan, line 15: have to update this description and the below accordingly

server/api/controllers/users/update-user.js
```diff
@@ -0,0 +1,24 @@
...
+    },
+    invalid : {
+      description: "Failed to retrieve user account",
+    }
+  },
+  fn: async function({ userId }, exits) {
+    User.updateOne(params.id, params)
```
albertparedandan, line 22: i'm guessing this file wasn't supposed to be in this PR, if so, can you remove it from this PR?

server/api/controllers/users/view-user.js
```diff
@@ -0,0 +1,41 @@
...
+    success: {
+      description: "User account returned successfully",
+    },
+    invalid : {
+      description: "Failed to retrieve user account",
+    }
+  },
+  fn: async function({ userId }, exits) {
+    if (userId) {
+      try {
```
albertparedandan, line 23: can refactor this into just 1 try/catch block, maybe to something like 
```
try {
  if (userId) {
    get user id
    if (data.length === 0) throw('user not found')
  }
  
    get all users
    if (data.length === 0) throw('no users')
} catch (err) {
  handle_error
}
```

server/config/routes.js
```diff
@@ -25,7 +25,8 @@ module.exports.routes = {
   "get /api/sermons/get-sermon-by-id": "sermons/get-sermon-by-id",
   "get /api/sermons/get-speaker": "sermons/get-speaker",
   "get /api/sermons/get-sermon-series": "sermons/get-sermon-series",
-  "post /api/users/add-user": "users/add-user",
+  
+  "get /api/users/view-user": "users/view-user",
```
albertparedandan, line 29: can we keep the naming convention consistent? since all of our other APIs are called `get*`, maybe we can also have `get-user`? unless there's a specific reason why we are using `view-user` that i'm not understanding

albertparedandan (changes requested):

server/api/controllers/auth/post-signup.js
tamarayustian: yeah good q. i thought about that too and thinking more like a drop down table for frontend. is it better if its a table of countries or an array of strings stored as constant?

server/config/routes.js
tamarayustian: not really any specific reason. i can change it to get-user

server/api/controllers/users/view-user.js
tamarayustian: actually now that i think about it, when the admin wants to find a user, they wouldn't use the userId. they would use emails or full names. so this query need to be adjusted to accommodate that right?

server/api/controllers/auth/post-signup.js
```diff
@@ -77,13 +84,15 @@ the account verification message.)`,
     emailAddress,
     password,
     fullName,
+    nationality,
     lifestage,
     phoneNumber,
   }) {
     if (
       emailAddress == null ||
```
albertparedandan, line 92: probs have to check for empty strings too

server/api/controllers/users/delete-user.js
```diff
@@ -0,0 +1,41 @@
...
+    missingRequiredFields: {
+      statusCode: 409,
+      description: "Please fill in the required fields.",
+    },
+  },
+
+  fn: async function({ userId }, exits) {
+    if (userId) {
+      try {
+        let data = await User.updateOne(userId).set({isDelete: true});
```
albertparedandan, line 30: does this handle if userId not found already?

server/api/controllers/users/update-user.js
```diff
@@ -0,0 +1,41 @@
...
+    missingRequiredFields: {
+      statusCode: 409,
+      description: "Please fill in the required fields.",
+    },
+  },
+
+  fn: async function({ params }, exits) {
+    if (params.id) {
+      try {
+        let data = await User.updateOne(params.id, params);
```
albertparedandan, line 30: same question as delete-user

server/api/controllers/users/delete-user.js
albertparedandan: also `isDeleted` :D

server/config/models.js
```diff
@@ -69,9 +69,10 @@ module.exports.models = {
   ***************************************************************************/
 
   attributes: {
-    createdAt: { type: 'number', autoCreatedAt: true, },
-    updatedAt: { type: 'number', autoUpdatedAt: true, },
+    createdAt: { type: 'ref', columnType: 'datetime', autoCreatedAt: true, },
+    updatedAt: { type: 'ref', columnType: 'datetime', autoUpdatedAt: true, },
     id: { type: 'number', autoIncrement: true, },
+    isDelete: { type: 'boolean', defaultsTo: false }
```
albertparedandan, line 75: isDeleted :D

albertparedandan (changes requested):

server/api/controllers/auth/post-signup.js
tamarayustian: nationality = null and nationality = "" is different?

albertparedandan: yep, `nationality == null` only checks for null and undefined

server/api/controllers/users/delete-user.js
```diff
@@ -0,0 +1,48 @@
...
+    },
+
+    missingRequiredFields: {
+      statusCode: 409,
+      description: "Please fill in the required fields.",
+    },
+  },
+
+  fn: async function ({ userId }, exits) {
+    if (userId) {
```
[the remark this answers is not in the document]
albertparedandan: this is much cleaner esp since we also do not need to throw "missingRequiredFields" since sails already handles that when you defined the input to be `required=true`

server/api/controllers/users/delete-user.js
```diff
@@ -0,0 +1,48 @@
...
+      try {
+        let data = await User.updateOne({ _id: userId, isDeleted: false }).set({
+          isDeleted: true,
+        });
+        if (data != null) {
+          return exits.success(data);
+        }
+        return exits.invalid();
+      } catch (err) {
+        if ((err.code = "E_CANNOT_INTERPRET_AS_OBJECTID")) {
```
albertparedandan, line 38: can we change this to just
```
catch(err) {
  sails.log.error(err);
  return exits.error(err);
}
```

server/api/controllers/users/update-user.js
```diff
@@ -0,0 +1,55 @@
...
+        delete params.id;
+        let data = await User.updateOne({ _id: userId, isDeleted: false }).set(
+          params
+        );
+        if (data != null) {
+          return exits.success(data);
+        }
+        return exits.invalid();
+      } catch (err) {
+        if ((err.code = "E_CANNOT_INTERPRET_AS_OBJECTID")) {
```
albertparedandan, line 45: same comments as `delete-user`.

server/api/controllers/users/update-user.js
```diff
@@ -0,0 +1,55 @@
...
+      statusCode: 409,
+      description: "The user id is invalid",
+    },
+  },
+
+  fn: async function ({ params }, exits) {
+    if (params.id) {
+      let userId = params.id;
+      try {
+        delete params.id;
```
albertparedandan, line 36: you could combine line 34 and line 36 using spread syntax:

```
const { id: userId, ...toUpdate } = params;
// userIdwill contain the id
// toUpdate will contain everything else except id
```

server/api/models/user.js
```diff
@@ -17,6 +17,10 @@ module.exports = {
       type: "string",
```
albertparedandan, line 17: models should start with capital letter (eg. User.js) so that we avoid confusion (all our ORM calls use capital letters)

server/config/keys.js
```diff
@@ -0,0 +1,13 @@
+const username = "root"
```
albertparedandan, line 1: this shouldn't be here

server/config/routes.js
```diff
@@ -36,6 +39,10 @@ module.exports.routes = {
   "post /api/auth/login": "auth/post-login",
   "post /api/auth/signup": "auth/post-signup",
   "post /api/auth/signup-google": "auth/post-signup-google",
+
```
albertparedandan, line 42: can you put a comment like the others just to help fix merge conflicts in the future

albertparedandan (changes requested):

tamarayustian:
added address field to user model
get-user can populate membership, baptism if info is there

PR #26, opened 2021-07-15
add refreshCache endpoint

albertparedandan:


server/api/controllers/cache/refresh-cache.js
```diff
@@ -0,0 +1,31 @@
...
+    nonSuccess: {
+      description: 'Error'
+    },
+  },
+
+  fn: async function(inputs, exits) {
+    sails.log.info(`Refreshing cache..`);
+
+    try {
+      sails.cache.flushAll();
```
tyleung, line 21: is there a possibility of some sort of race condition issue here? If so, setting cacheLatest() to a tmp variable before doing a flush might be better

albertparedandan: true. Might have to look into more robust caching libraries that allows locking (currently node-cache which is what we use doesn't support lock) but it will get more complicated. We decided to accept this risk of race condition because worst case scenario is the server just takes a while to load due to cache being empty during this flushing period.

tyleung (approved):

PR #29, opened 2021-07-17
13 feature form creation system

ksuhartono97:


server/api/controllers/forms/get-form-by-id.js
```diff
@@ -0,0 +1,25 @@
+module.exports = {
```
albertparedandan, line 1: i think we should combine `get-all-forms` with `get-form-by-id` to just 1 controller: `get-forms` where id is optional, if no id given, return all. it will make it much cleaner and it is also the actual implementation of `Model.find()` anyway. Can take a look at other controllers (sermons, announcements, posts etc.) if need more example

server/api/controllers/forms/post-create-form.js
```diff
@@ -0,0 +1,32 @@
...
+      description: "New form was created successfully.",
+    },
+    error: {
+      description: "Failed to create new form.",
+    },
+  },
+
+  fn: async function ({ formToSave }, exits) {
+    try {
+      const newForm = await Form.create(formToSave).fetch();
```
albertparedandan, line 25: hmm, why do we need `.fetch()` if we are not using newForm for anything else afterwards anyway?

server/api/controllers/forms/post-create-form.js
```diff
@@ -0,0 +1,32 @@
...
+      return exits.success();
```
albertparedandan, line 26: also, wondering if we need to check the JSON object to make sure there's nothing wrong with it before saving?

server/api/controllers/forms/post-delete-form.js
```diff
@@ -0,0 +1,25 @@
...
+      required: true,
+      description: "Id of form",
+    },
+  },
+
+  exits: {},
+
+  fn: async function ({ id }, exits) {
+    try {
+      const data = await Form.destroy(id);
```
albertparedandan, line 18: i'm assuming this already handles if id doesn't exist? if not, then will this already handle that situation?

server/api/controllers/forms/post-update-form.js
```diff
@@ -0,0 +1,30 @@
...
+      required: true,
+      description: "Updated JSON data",
+    },
+  },
+
+  exits: {},
+
+  fn: async function ({ id, formToSave }, exits) {
+    try {
+      const updatedForm = await Form.updateOne(id).set(formToSave);
```
albertparedandan, line 23: same comment about handling non-existent id if not yet handled

server/api/controllers/forms/post-delete-form.js
albertparedandan: also, should probably consider using soft-delete

server/api/models/Form.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  attributes: {
+    formName: {
+      type: "string",
+      required: true,
+    },
+    formFields: {
+      type: "json",
+    },
+  },
```
albertparedandan, line 10: soft-delete

albertparedandan (changes requested):

server/config/routes.js
```diff
@@ -20,6 +20,13 @@ module.exports.routes = {
 
   "get /api/cache/refresh-cache": "cache/refresh-cache",
 
+  "post /api/forms/create-form": "forms/create-form",
```
albertparedandan, line 23: mayb arrange this a bit neater (use comments, group gets and posts together etc.) to help solve merge conflicts in the future

ui/src/components/MainContainer.js
```diff
@@ -7,6 +7,7 @@ import { MuiPickersUtilsProvider } from "material-ui-pickers";
 import SermonContainer from "./sermons/SermonContainer";
```
albertparedandan, line 7: assuming all UI changes below will be modified again and not final

albertparedandan (changes requested):

server/api/controllers/forms/post-create-form.js
ksuhartono97: yea this will be handled in future with validation which isn't currently implemented yet

ui/src/components/MainContainer.js
ksuhartono97: yup that's temporary only

server/config/routes.js
ksuhartono97: yea this line is actually a merge conflict. it shouldn't be there

albertparedandan (approved):
nice

PR #70, opened 2021-08-15
GH-58: event card fixes so it doesn't break on safari

ksuhartono97:


ui/src/components/events/EventCard.js
```diff
@@ -117,33 +117,44 @@ const EventCard = (props) => {
...
-          >
-            Add to Calendar
-          </Button>
+          {eventData.signUpLink.length > 0 && (
+            <Button
+              flex={1}
+              as={Link}
+              size="sm"
+              target="_blank"
+              href={eventData.signUpLink ? eventData.signUpLink : false}
```
albertparedandan, line 139: this returns error. you cannot pass a boolean to href. return null instead. same for your other buttons and your generateGoogleCalendarLink

albertparedandan (changes requested):

ui/src/components/events/EventCard.js
albertparedandan: also, i'll prob do it in my homepage branch, but i think a lot of this logic should be in our backend

ksuhartono97: yea, if it is available directly from our backend it makes a lot of this so much easier and for frontend it should just be getting data from BE. temporary workaround for now since that involves editing the advanced custom fields

ui/src/components/events/EventsPage.js
```diff
@@ -21,27 +21,31 @@ const EventsPage = (props) => {
...
+              item.recurrence
+            );
+            item.renderDate = renderDate;
+            return endDate > DateTime.now();
+          } else return false;
+        });
+        filtered.sort((a, b) => (a.renderDate > b.renderDate ? 1 : -1));
+        setEventsList([...filtered]);
+      } else {
+        throw Error("Something went wrong with the requesy")
```
albertparedandan, line 40: lol typo

albertparedandan (approved):

albertparedandan (approved):

PR #71, opened 2021-08-16
DB Field changes for countryOfOrigin and Address as object

tamarayustian:
- changing nationality to countryOfOrigin because it makes more sense for users to put in where they're from, rather than what their nationality is (a less common term)
- changing address to a json type, so user can input their full address and we can store it more clearly. so address from now on will be:
`address: { flat, floor, street, district }`


ksuhartono97 (approved):
looks good to me

PR #109, opened 2021-10-30
GH-108: implementation of no login user forms

### LINKED ISSUE #108: Forms: Non-login user form

ksuhartono97:
Closes #108

### DIFF
--- a/server/api/controllers/forms/get-form.js
+++ b/server/api/controllers/forms/get-form.js
@@ -9,17 +9,33 @@ module.exports = {
       type: "string",
       description: "Id of form",
     },
+    isLoggedIn: {
+      type: "boolean",
+      description: "whether the user is logged in or not",
+      required: true,
+    },
   },
 
   exits: {},
 
-  fn: async function ({ id }, exits) {
+  fn: async function ({ id, isLoggedIn }, exits) {
     try {
-      const data = await Form.find({
+      let searchParams = {
         _id: id,
         isPublished: true,
         isDeleted: false,
-      });
+      };
+
+      // If user is not logged in, only allow them to retrieve non login forms
+      if (!isLoggedIn) {
+        searchParams["requireLogin"] = false;
+      }
+
+      const data = await Form.find(searchParams);
+      
+      // If no form is found return error
+      if (data === null) return exits.error("unauthorized access");
+
       return exits.success(data);
     } catch (err) {
       sails.log(err);
--- a/server/api/controllers/forms/get-form.js
+++ b/server/api/controllers/forms/get-form.js
@@ -20,19 +20,13 @@ module.exports = {
 
   fn: async function ({ id, isLoggedIn }, exits) {
     try {
-      let searchParams = {
+      const data = await Form.find({
         _id: id,
         isPublished: true,
         isDeleted: false,
-      };
+        requireLogin: !isLoggedIn ? isLoggedIn : [false, true],
+      });
 
-      // If user is not logged in, only allow them to retrieve non login forms
-      if (!isLoggedIn) {
-        searchParams["requireLogin"] = false;
-      }
-
-      const data = await Form.find(searchParams);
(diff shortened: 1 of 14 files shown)


server/api/controllers/forms/get-form.js
```diff
@@ -9,17 +9,33 @@ module.exports = {
...
-  fn: async function ({ id }, exits) {
+  fn: async function ({ id, isLoggedIn }, exits) {
     try {
-      const data = await Form.find({
+      let searchParams = {
         _id: id,
         isPublished: true,
         isDeleted: false,
-      });
+      };
```
albertparedandan, line 27:
```
searchParams = {
  _id: id,
  isPublished = true,
  isDeleted = false,
  requireLogin = !isLoggedIn
};
```

server/api/controllers/forms/post-create-submission.js
```diff
@@ -75,21 +81,29 @@ module.exports = {
...
+        email = submissionData["email"];
+        fullName = submissionData["fullName"];
+      }
+
+      // Send confirmation email if there is email
+
+      if (email) {
         await sails.helpers.sendTemplateEmail.with({
-          to: user.email,
+          to: email,
```
albertparedandan, line 95: can do
```
to: user.email ? user.email : submissionData["email"]
fullName: user.fullName ? user.fullName : submissionData["fullName"]
```

and remove line 44-46, 84-92

ui/src/components/forms/FormManager.js
```diff
@@ -64,12 +74,9 @@ const FormManager = (props) => {
         );
       }
 
-      setValue("formName", data[0].formName);
-      setValue("formDescription", data[0].formDescription);
-      setValue("formImage", data[0].formImage);
-      setFormName(data[0].formName);
-      setFormDescription(data[0].formDescription);
-      setFormImage(data[0].formImage);
+      console.log(data);
```
albertparedandan, line 77: remove pls

ui/src/components/helpers/userInformationHelpers.js
```diff
@@ -86,7 +86,10 @@ const updateUserDataRequest = async (data) => {
 };
 
 const getPublicFormsRequest = async () => {
```
albertparedandan, line 88: is this properly named? my thought of public is actually is isLoggedIn = false cus public means anyone can access right

albertparedandan (changes requested):

server/api/controllers/forms/get-form.js
ksuhartono97: no i realized can't do this. if we do this then it would only return forms where `requireLogin = !isLoggedIn`

if `isLoggedIn` is `true` then we want to get all forms whether their `requireLogin` is `true` or `false`. But when `isLoggedIn` is `false` then we should only return the forms where `requireLogin` is `false`. Unless I'm not getting how the params search works?

albertparedandan: Ah ok I get what you mean

albertparedandan: Fine to leave it as it is now. Wondering if null works?

```
requireLogin: !isLoggedIn ? isLoggedIn : null
```

albertparedandan (approved):

server/api/controllers/forms/get-form.js
```diff
@@ -9,17 +9,27 @@ module.exports = {
...
   exits: {},
 
-  fn: async function ({ id }, exits) {
+  fn: async function ({ id, isLoggedIn }, exits) {
     try {
       const data = await Form.find({
         _id: id,
         isPublished: true,
         isDeleted: false,
+        requireLogin: !isLoggedIn ? isLoggedIn : [false, true],
```
albertparedandan, line 27: smart

albertparedandan (approved):

PR #114, opened 2021-11-25
connected form manager to adminform

tamarayustian:


ui/src/components/admin/AdminForm.js
```diff
@@ -1,19 +1,9 @@
...
-        bg={useColorModeValue("gray.200")}
-        justifyContent="center"
-      >
-        <Flex direction="column" p={12} rounded={6}>
-          <Heading mb={6} alignItems="center">
-            Forms
-          </Heading>
-        </Flex>
-      </Flex>
+      <FormManager />
```
albertparedandan, line 6: hmm, code wise is fine. but don't particularly like the fact that when going to /admin/forms without logging in, it shows a 404 instead of redirect to login form

albertparedandan (approved):

PR #213, opened 2022-02-12
188 content restriction

ksuhartono97:


server/api/controllers/whitelist/get-whitelist.js
```diff
@@ -0,0 +1,27 @@
+module.exports = {
+  friendlyName: 'Get whitelist',
+
+  description:
+    'Retrieves the list of whitelists that exist if id is not specified, finds a specific whitelist entry otherwise',
+
+  inputs: {
+    id: {
+      type: 'string',
```
albertparedandan, line 9: do we have to put `required: false` here?

server/config/routes.js
```diff
@@ -19,70 +19,75 @@ module.exports.routes = {
...
+
+  // Whitelist APIs
+  'get /api/whitelist/get': 'whitelist/get-whitelist',
+  'post /api/whitelist/create': 'whitelist/post-create-whitelist',
+  'post /api/whitelist/update': 'whitelist/post-update-whitelist',
 
   // UI
-  "get /*": {
+  'get /*': {
     skipAssets: true,
```
albertparedandan, line 88: actually, is this the reason why we cannot access assets in our server lol, like the brp stuff have to be in images folder

ui/src/components/admin/AdminWhitelistManager.js
```diff
@@ -0,0 +1,64 @@
...
+      <Container maxW="container.lg" minHeight="85vh" p={5}>
+        <Heading mb={6} alignItems="center">
+          Whitelist Manager
+        </Heading>
+        <Text>
+          Insert whitelist here (hardcoded for retreat, don't touch if you don't
+          know what you're doing)
+        </Text>
+        <form onSubmit={handleSubmit(submitWhitelist)}>
+          <Textarea
```
albertparedandan, line 49: do we not want to populate this with the whitelist if it already exists in the mongo?

albertparedandan (approved):
no major comments, just suggestions

ui/src/components/admin/AdminWhitelistManager.js
ksuhartono97: don't think this is necessary, the interface is there to simply just quickly update if needed haha

server/config/routes.js
ksuhartono97: i dont think so actually lolol

PR #393, opened 2022-05-17
392 content restriction

tamarayustian:
hardcoded for ignite right now

### DIFF
--- a/server/api/helpers/auth/verify-jwt.js
+++ b/server/api/helpers/auth/verify-jwt.js
@@ -1,33 +1,33 @@
-const jwt = require("jsonwebtoken");
+const jwt = require('jsonwebtoken');
 
 module.exports = {
-  friendlyName: "Verify JWT",
-  description: "Verify a JWT token.",
+  friendlyName: 'Verify JWT',
+  description: 'Verify a JWT token.',
   inputs: {
     token: {
-      type: "string",
-      friendlyName: "token",
-      description: "A reference to the request object (req).",
+      type: 'string',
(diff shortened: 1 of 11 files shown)


ui/src/components/admin/AdminWhitelistManager.js
```diff
@@ -0,0 +1,92 @@
...
+            status: 'success',
+            duration: 3000,
+            isClosable: true,
+          });
+        }
+        reset();
+      } else {
+        let igniteEntry = data.filter((obj) => obj.eventName === 'ignite-2022');
+        if (igniteEntry.length === 1) {
+          let { status } = await axios.post('/api/whitelist/update', {
```
bhryck, line 38: when testing, updating the whitelist overwrites the past records. not sure if this is intended behaviour

tamarayustian: yeah that is the intended behaviour

server/api/helpers/auth/verify-jwt.js
```diff
@@ -37,10 +37,16 @@ module.exports = {
         ...result
       } = user;
 
+      const whitelistData = await Whitelist.findOne({
+        eventName: 'ignite-2022',
+        data: { contains: valid.emailAddress },
+      });
+      result.whitelisted = whitelistData ? true : false;
```
albertparedandan, line 44: since this is not going into release branch, i think it's fine to just leave it, but this should be a bit more descriptive, e.g. igniteWhitelisted

PR #421, opened 2022-06-24
GH-407: added t3ch access type option

tamarayustian:


albertparedandan:
hmm actually, looking at this, it's probably better for t3ch access to have their own branch.

reasoning is because if let's say i approve this n gets merged into release branch, release branch will have a role 't3ch' that may not be handled properly, which at the worst case scenario, means release branch isn't stable.

by creating an accessType branch (where all future additional implementations of the access type will be merged into), it ensures that 1. smaller PRs to review, 2. release branch is always stable and can run

TLDR: all other PRs related to accessType should request merge to `407-t3ch-access-type`, be reviewed before merged, and then this final PR will be merged into release once ready

let me know if unclear

albertparedandan (approved):

PR #428, opened 2022-06-26
408 secure policy

Baeck016:
Done with cache policy


server/config/policies.js
```diff
@@ -36,7 +36,7 @@ module.exports.policies = {
   "forms/post-contact-us-form": true,
 
   // Cache
-  "cache/*": ["isLoggedIn", "isAdmin"],
+  "cache/*": ["isLoggedIn", "isTech"],
```
bhryck, line 39: https://hongkong.hmcc.net/clear-cache is the current link to clear cache in the website. so should change this to /clear-cache

bhryck, line 39: since isTech.js is actually checking if the accestype is "t3ch", "admin" or "stewardship", it is best to rename the function to reflect that. for example "isTechOrAbove". otherwise we might mistake it for checking if accestype is "tech" only.

albertparedandan (approved):

server/config/policies.js
albertparedandan: nice

PR #430, opened 2022-06-30
GH-414: Directly update empty user information via form submission

ksuhartono97:
Handled both null and "" because sometimes the value cannot be set to null due to it being a required field


albertparedandan:
Should this PR be request to merge to your other branch instead?

albertparedandan (approved):

PR #473, opened 2022-07-24
Release

albertparedandan:


ui/src/components/NavigationBar/NavBar.js
```diff
@@ -73,7 +74,7 @@ const NavBar = (props) => {
 
   return (
     <>
-      <Flex w="100vw" background="rgba(0, 0, 0, 0.4)" justify="center">
+      <Flex w="100vw" background="rgba(0, 0, 0, 0.4)" justify="center" display={ isAdminPage ? 'none' : 'block' }>
```
ksuhartono97, line 77: dk if it's necessary to do this, i think i would still want to navigate out of the admin page using the navbar

ksuhartono97 (changes requested):

ksuhartono97 (approved):

PR #478, opened 2022-08-02
452 error pages

ksuhartono97:


ui/src/components/screens/ErrorPage.js
```diff
@@ -0,0 +1,150 @@
...
+          return 'form-unavailable';
+        case '/form-will-open':
+          return 'form-will-open';
+        case '/form-success':
+          return props.user.id ? 'form-success-logged-in' : 'form-success';
+        default:
+          return 'not-found';
+      }
+    } catch (err) {
+      console.error('Error detected in error page paths');
```
bhryck, line 79: [Bug] when logging in, the not-found error page shows up for half a second before the page renders. i suspect is because of this snippet

ui/src/components/screens/ErrorPage.js
```diff
@@ -0,0 +1,150 @@
...
+          return 'need-login';
+        case '/need-fill-profile':
+          return 'need-fill-profile';
+        case '/form-unavailable':
+          return 'form-unavailable';
+        case '/form-will-open':
+          return 'form-will-open';
+        case '/form-success':
+          return props.user.id ? 'form-success-logged-in' : 'form-success';
+        default:
```
bhryck, line 75: [Bug] when logging in, the not-found error page shows up for half a second before the page renders. i suspect is because of this snippet

ksuhartono97: behavior was similar since before, it always shows the not-found page for a bit when logging in because of how our login behavior works. not sure how to handle it?

albertparedandan (approved):

Issue #459, opened 2022-07-18
Auth button behaviours and design change


tamarayustian:
Questions:
- Is connect and event menu flipped? or the design change is intended?
- Is there a specific behavior on HMCC title?
![image](https://user-images.githubusercontent.com/26876443/179533758-f3468995-810d-4a75-a86d-9bc7cff9b289.png)

PR #481, opened 2022-08-07
GH-459: navbar

Baeck016:
Done with navbar


ui/src/components/NavigationBar/MainMenu.js
```diff
@@ -173,9 +172,17 @@ const MainMenu = ({ login, onClose }) => {
...
                     </LinkOverlay>
                   </Button>
+                  <Button
+                    onClick={onClose}
+                    marginBottom="5"
+                    width="100%"
+                    style={{ WebkitMarginStart: '0', marginInlineStart: '0' }}
+                  >
+                    <LinkOverlay href="/signup">SIGN UP</LinkOverlay>
+                  </Button>
```
bhryck, line 185: image
the signup button should be the same color as the navbar when not hovered on.

bhryck: actually saw diff versions on the figma, if that the one cxo decided on then no need to change

ui/src/components/NavigationBar/MainMenu.js
```diff
@@ -137,17 +137,16 @@ const MainMenu = ({ login, onClose }) => {
                 >
```
bhryck, line 137: image
image

color of the bg looks different, may need to check with cxo if they want to change the bg image color

ui/src/components/NavigationBar/MainMenu.js
Baeck016: > actually saw diff versions on the figma, if that the one cxo decided on then no need to change

It's the one top, the one with blue and dark blue when hovered on.

package.json
```diff
@@ -5,6 +5,10 @@
     "prettier": "2.3.2"
   },
   "dependencies": {
+    "ag-grid-community": "^28.0.2",
```
albertparedandan, line 8: this isn't supposed to be here. please remove

albertparedandan (changes requested):

ui/src/components/NavigationBar/MainMenu.js
```diff
@@ -137,17 +137,16 @@ const MainMenu = ({ login, onClose }) => {
...
                   <Button marginBottom="5" width="100%">
                     <LinkOverlay href="/profile" onClick={onClose}>
-                      My Profile
+                      MY PROFILE
```
albertparedandan, line 140: all these buttons do not work when i press the actual button itself. have to click the text.

albertparedandan (approved):

PR #482, opened 2022-08-07
465 restyling visit page

cetair:


ui/src/components/visit-us/VisitUsPage.js
```diff
@@ -175,7 +175,7 @@ const VisitUsPage = (props) => {
               textAlign={['center', 'right']}
             >
               <Link href="https://hongkong.hmcc.net/online/" isExternal>
-                Join us Online <Icon as={RiArrowRightLine} />
+                Join us ONLINE
```
albertparedandan, line 178: this is not removed in the figma demo

ui/src/components/visit-us/VisitUsPage.js
```diff
@@ -175,7 +175,7 @@ const VisitUsPage = (props) => {
...
               </Link>
             </Heading>
           </VStack>
```
albertparedandan, line 181: also, i think the FAQ in the figma is different. can you double check if we're supposed to change the FAQs too?

albertparedandan (changes requested):

albertparedandan (approved):

Issue #454, opened 2022-07-18
Auto publish & take down forms


ksuhartono97:
Low priority. Raised by CQO

ksuhartono97:
Forms can be published but should not render form contents if it is not within the defined form active time (should still work if date is not always shown)

ksuhartono97:
The process:
1. Add Form Available From & Form Available Until to the saved form data in the backend
2. Handle the logic of making form available/unavailable in the front end

PR #493, opened 2022-08-17
GH-454: Auto publish & takedown forms

ksuhartono97:


ui/src/components/forms/UserFormContainer.js
```diff
@@ -23,6 +24,34 @@ const UserFormContainer = (props) => {
...
+            // is valid: check if it's within the time of form availability
+            // is not valid: default to whatever value was already there
+            let beforeEndTime = formAvailableUntil.isValid
+              ? DateTime.now() < formAvailableUntil
+              : true;
+            let isInRange = afterStartTime && beforeEndTime;
+
+            if (!isInRange) {
+              if (!beforeEndTime) history.push('/form-unavailable');
+              else history.push('/form-will-open');
```
albertparedandan, line 52: Should we put form open time in the error page?

ui/src/components/forms/UserFormContainer.js
```diff
@@ -23,6 +24,34 @@ const UserFormContainer = (props) => {
...
+          const formAvailableUntil = DateTime.fromISO(
+            data[0].formAvailableUntil
+          ).setZone('Asia/Hong_Kong');
+
+          // One of these values are set
+          if (formAvailableFrom.isValid || formAvailableUntil.isValid) {
+            // If form available from is set & valid, check if current time is
+            // after formAvailableFrom
+            let afterStartTime = formAvailableFrom.isValid
+              ? DateTime.now() >= formAvailableFrom
```
albertparedandan, line 39: Should this be DateTime.local()?

ui/src/components/forms/UserFormContainer.js
```diff
@@ -23,6 +24,34 @@ const UserFormContainer = (props) => {
         // Form is not published, hence it is unavailable
         if (!data[0]) history.push('/form-unavailable');
         else {
+          const formAvailableFrom = DateTime.fromISO(
+            data[0].formAvailableFrom
+          ).setZone('Asia/Hong_Kong');
+          const formAvailableUntil = DateTime.fromISO(
```
albertparedandan, line 30: Should this be DateTime.fromISO().toLocal()?

ui/src/components/forms/UserFormContainer.js
ksuhartono97: see other comment

ui/src/components/forms/UserFormContainer.js
```diff
@@ -23,6 +24,38 @@ const UserFormContainer = (props) => {
...
+            let beforeEndTime = formAvailableUntil.isValid
+              ? DateTime.now().setZone('Asia/Hong_Kong') < formAvailableUntil
+              : true;
+            let isInRange = afterStartTime && beforeEndTime;
+
+            if (!isInRange) {
+              if (!beforeEndTime) history.push('/form-unavailable');
+              else
+                history.push({
+                  pathname: '/form-will-open',
```
albertparedandan, line 54: hmmm. have a feeling people will complain about this. since we are redirecting to an error page, refreshing the page won't bring them back to this form. i personally sometimes go to a form at like 7:59 and refresh until it opens up

ui/src/components/forms/UserFormContainer.js
```diff
@@ -1,7 +1,8 @@
 import React, { useState, useEffect } from 'react';
 import { customAxios as axios } from '../helpers/customAxios';
 import Form from './Form';
-import { Container, Center, VStack, Text } from '@chakra-ui/react';
+import { Container } from '@chakra-ui/react';
+import { DateTime } from 'luxon';
 
 const UserFormContainer = (props) => {
   const [formData, setFormData] = useState(null);
```
albertparedandan, line 8: there is a split second where the form loads and then redirects to an error page. i think we need to set a `isLoading` variable here to only render the form once isLoading is set to true (at the end of your useEffect).

ui/src/components/forms/UserFormContainer.js
albertparedandan: hmm, i tried both. both still relies on local device's time settings. so changing the date to the future will unlock the form. i think the only way to solve this is to use our server's datetime instead of client's datetime

albertparedandan:
Approved on the assumption that comments will be addressed as future enhancements

albertparedandan (approved):

Issue #495, opened 2022-08-18
Connect page testimonies update


tamarayustian:
Design:
![image](https://user-images.githubusercontent.com/26876443/185450207-49b8cdaf-0376-411f-92f2-a64dba98a46d.png)

Photos:
https://drive.google.com/drive/folders/1AOIzVdix0dWehnQ2irj9trtfTKtGQa32?usp=sharing

IG links:
1. josephine - https://www.instagram.com/p/ChPKak1vz9L/
2. aurora - https://www.instagram.com/p/ChPK7GWvYgt/
3. dr yap- https://www.instagram.com/p/ChPLhgTPdkc/
4. chelsea - https://www.instagram.com/p/ChR_dFPvSPE/
5. track - https://www.instagram.com/p/ChUZYTpvGc7/
6. seivabel - https://www.instagram.com/p/ChUc0O3PLoy/
7. yuen yau - https://www.instagram.com/p/ChXddPnvM_O/
8. david - https://www.instagram.com/p/ChZecLJPGEm/
9. chloe - https://www.instagram.com/p/Chb6VtvPjYc/
10. vimbai -

PR #498, opened 2022-08-20
GH-495: update testimonies for IRL

tamarayustian:


albertparedandan (approved):

albertparedandan:
hmm, i wonder if this will create conflict with Cath's connect page

albertparedandan:
also, are those images already compressed? given the small size in the browser it's quite big

tamarayustian:
> 

actually not compressed yet. but i can only do it at night

Issue #457, opened 2022-07-18
1st Half of homepage (vision, mission, sermon, main image)


tamarayustian:
![image](https://user-images.githubusercontent.com/26876443/179687874-db930d98-99d9-42b6-87b7-767a2e891d4f.png)

PR #502, opened 2022-08-21
GH-457: Video will play after loaded, if not stay at 00:00

Midnite1212:


ui/src/components/home/HelloSermonSection.js
```diff
@@ -24,7 +24,7 @@ const HelpCardInfo = [
   },
   {
     title: 'Visit us',
-    text: ['Every Sunday at 10AM HKT',<br />, 'In-person & Online'],
+    text: ['Every Sunday at 10AM HKT',<br key="space" />, 'In-person & Online'],
```
albertparedandan, line 27: what is this key for ya?

Midnite1212: To fix warning key in console

albertparedandan (approved):

Issue #522, opened 2022-09-25
Events - show from/to dates


albertparedandan:
- enable the field in WordPress
- pass the field to webapp
- show in UI

PR #555, opened 2022-10-08
GH-522: Show StartDate and EndDate in Events

albertparedandan:


ui/src/components/home/EventsSectionCards.js
```diff
@@ -131,19 +145,27 @@ const EventsSectionCard = (props) => {
...
+                    {event.renderDate
+                      ? event.renderDate.toLocaleString(
+                          DateTime.DATE_MED_WITH_WEEKDAY
+                        )
+                      : getRenderDate(
+                          event.startDate,
+                          event.endDate,
+                          event.recurrence
+                        ).toLocaleString(DateTime.DATE_MED_WITH_WEEKDAY)}
+                  </Text>
```
bhryck, line 161: would it be better to move this to eventshelper like EndDateElement

albertparedandan: refactored

PR #570, opened 2022-10-12
539 migrate online page

Midnite1212:


package-lock.json
```diff
@@ -0,0 +1,329 @@
+{
```
albertparedandan, line 1: why is this changed?

albertparedandan: same question for all the package.json and yarn.lock files

ui/env
```diff
@@ -0,0 +1,2 @@
+REACT_APP_GOOGLE_CLIENT_ID=694013380015-j2lgv2uucaqs6shl72km6unmq77v74fn.apps.googleusercontent.com
```
albertparedandan, line 1: why is this changed?

server/config/routes.js
```diff
@@ -40,6 +40,11 @@ module.exports.routes = {
   // Pages APIs
   'get /api/pages/get-pages': 'pages/get-pages',
 
+  //Live Semron APIs
+  'post /api/live-sermons/create-live-sermon': 'liveSermon/create-live-sermon',
```
albertparedandan, line 44: can we call this live-sermon instead of live-sermons? keeping in consistent

albertparedandan: same comment for the others

server/api/controllers/liveSermon/create-live-sermon.js
```diff
@@ -0,0 +1,103 @@
...
+    sermonPassage: {
+      type: 'string',
+    },
+    sermonDescription: {
+      type: 'string',
+    },
+    isPublished: {
+      type: 'boolean',
+    },
+    lastUpdatedBy: {
```
albertparedandan, line 38: i would remove this. and just always assign lastUpdatedBy in your function implementation

server/api/controllers/liveSermon/update-live-sermon.js
```diff
@@ -0,0 +1,123 @@
...
+    sermonPassage: {
+      type: 'string',
+    },
+    sermonDescription: {
+      type: 'string',
+    },
+    isPublished: {
+      type: 'boolean',
+    },
+    lastUpdatedBy: {
```
albertparedandan, line 42: similar comment on create-live-sermon. i would remove this

server/api/controllers/liveSermon/update-live-sermon.js
```diff
@@ -0,0 +1,123 @@
...
+        sermonDate,
+        sermonPassage,
+        sermonDescription,
+        isPublished,
+        lastUpdatedBy,
+        isDeleted,
+      });
+
+      if (!existing) {
+        throw `Name: ${title}, live sermon doesn't exist`;
```
albertparedandan, line 111: would show the sermon id too in the logs

albertparedandan:
there's a lot of commits here that do not belong. can you remove them?

albertparedandan (changes requested):

package-lock.json
Midnite1212: i think it got updated to the latest version when I do yanr install. should I revert it?

ui/env
Midnite1212: eh.. why is the .env uploaded, it should be in the .gitignore? ?

package-lock.json
albertparedandan: yes pls

albertparedandan:
What happened to isPublished variable? I think it's useful so we should keep it

albertparedandan:
Also, please resolve comments so I know which ones you addressed

albertparedandan (changes requested):
See comments

Midnite1212:
ohh I thought in the prev comments u meant for last updated and isPublished, my bad will add it back

albertparedandan (approved):

PR #576, opened 2022-10-14
Hg instructions

Baeck016:
Done with the instructions, but haven't filled in the materials, and points.

### DIFF
--- a/ui/src/components/harvest-games/HarvestGameInstructions.js
+++ b/ui/src/components/harvest-games/HarvestGameInstructions.js
@@ -10,7 +10,8 @@ import {
   HStack,
   Grid,
 } from '@chakra-ui/react';
-import { useState } from 'react';
+import { Text } from '@chakra-ui/layout';
+import { useState, Fragment } from 'react';
 
 import './styles.css';
 
@@ -21,7 +22,8 @@ export default function HarvestGameInstructions() {
       name: 'SMUGGLING THROUGH SECURITY',
       round: 3,
       player: 5,
-      points: 15,
+      points: 45,
+      materials: '-',
       imageUrl: [
         'https://hongkong.sub.hmcc.net/wp-content/uploads/hg1_1.gif',
         'https://hongkong.sub.hmcc.net/wp-content/uploads/hg1_2.gif',
@@ -34,7 +36,18 @@ export default function HarvestGameInstructions() {
       game_num: 'Game #2',
       name: 'THE PACKING GAME',
       round: 2,
-      player: '-',
+      player: 2,
+      points: 230,
+      materials: [
+        '10 Shirts',
+        '6  SweatPants',
+        '6  Books',
+        '6  Hoodies',
+        '9  Socks',
+        '6  Shoes',
+        '8  Hats',
+        '6  Water bottles',
+      ],
       imageUrl: [
         'https://hongkong.sub.hmcc.net/wp-content/uploads/hg2_1.gif',
         'https://hongkong.sub.hmcc.net/wp-content/uploads/hg2_2.gif',
@@ -49,7 +62,9 @@ export default function HarvestGameInstructions() {
       game_num: 'Game #3',
       name: 'THE AIRPLANE RELAY RACE',
       round: 4,
-      player: '-',
+      player: 5,
+      materials: '-',
+      points: '-',
       imageUrl: [
         'https://hongkong.sub.hmcc.net/wp-content/uploads/hg3_1.gif',
         'https://hongkong.sub.hmcc.net/wp-content/uploads/hg3_2.gif',
@@ -131,10 +146,13 @@ export default function HarvestGameInstructions() {
       {gameType[selected].pwd === localStorage.getItem(gameType[selected].id) ||
       gameType[selected].pwd === password ? (
         <Box w="90%">
-          <Heading size="xl" textColor="rgba(83, 83, 83, 1)">
+          <Heading size="xl" textColor="rgba(1, 43, 85, 1)">
             {gameType[selected].game_num}
           </Heading>
-          <Heading size="lg" textColor="rgba(83, 83, 83, 1)">
+          <Heading
+            fontSize={{ base: '4vw', md: '2vw' }}
+            textColor="rgba(1, 43, 85, 1)"
+          >
             {gameType[selected].name}
           </Heading>
           <HStack
@@ -188,7 +206,7 @@ export default function HarvestGameInstructions() {
                 fontWeight={500}
                 fontSize={{ base: '4xl', md: '5xl' }}
               >
-                3
+                {gameType[selected].player}
               </Heading>
               <Heading
                 fontSize={{ base: 'sm' }}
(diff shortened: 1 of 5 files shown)


ui/src/components/harvest-games/HarvestGameInstructions.js
```diff
@@ -0,0 +1,357 @@
...
+import './styles.css';
+
+export default function HarvestGameInstructions() {
+  const gameType = [
+    {
+      game_num: 'Game #1',
+      name: 'SMUGGLING THROUGH SECURITY',
+      round: 3,
+      player: 5,
+      points: 45,
```
ksuhartono97, line 25: Actually can we remove the points, I believe we cannot actually give this as a number since the scoring is based on an internal system right

ui/src/components/harvest-games/HarvestGameInstructions.js
```diff
@@ -0,0 +1,357 @@
...
+      id: 'game_1',
+      pwd: 'Smuggling',
+    },
+
+    {
+      game_num: 'Game #2',
+      name: 'THE PACKING GAME',
+      round: 2,
+      player: 2,
+      points: 230,
```
ksuhartono97, line 40: remove points

ui/src/components/harvest-games/HarvestGameInstructions.js
```diff
@@ -0,0 +1,357 @@
...
+      pwd: 'Packing',
+    },
+
+    {
+      game_num: 'Game #3',
+      name: 'THE AIRPLANE RELAY RACE',
+      round: 4,
+      player: 5,
+      materials: '-',
+      points: '-',
```
ksuhartono97, line 67: remove points

ui/src/components/harvest-games/HarvestGameInstructions.js
```diff
@@ -0,0 +1,357 @@
...
+          >
+            {gameType[selected].name}
+          </Heading>
+          <HStack
+            justifyContent={('space-evenly', 'space-between')}
+            paddingY="5"
+            w="100%"
+            display="flex"
+          >
+            <Box
```
ksuhartono97, line 164: make conditional rendering, only render the necessary ones if the value is defined

ui/src/components/harvest-games/HarvestGameInstructions.js
```diff
@@ -0,0 +1,357 @@
...
+  const gameType = [
+    {
+      game_num: 'Game #1',
+      name: 'SMUGGLING THROUGH SECURITY',
+      round: 3,
+      player: 5,
+      points: 45,
+      materials: '-',
+      imageUrl: [
+        'https://hongkong.sub.hmcc.net/wp-content/uploads/hg1_1.gif',
```
ksuhartono97, line 28: i guess this means that all the images are hardcoded in beforehand? so the team has to provide all this before sat night

ui/src/components/harvest-games/HarvestGameInstructions.js
```diff
@@ -0,0 +1,357 @@
...
+  ];
+
+  const [selected, setSelected] = useState(0);
+  const [password, setPassword] = useState('');
+
+  const handleSubmit = () => {
+    let pwd = document.getElementById('inputted').value;
+    setPassword(pwd);
+    localStorage.setItem(gameType[selected].id, pwd);
+    console.log(localStorage.setItem(gameType[selected].id, pwd));
```
ksuhartono97, line 85: remove `console.log` like this

ui/src/components/harvest-games/HarvestGameInstructions.js
```diff
@@ -0,0 +1,357 @@
...
+            <Grid
+              w="100%"
+              mt="12"
+              mb="12"
+              templateColumns="repeat(1, 1fr)"
+              gap="2"
+            >
+              {gameType[selected].imageUrl.length > 0 &&
+                gameType[selected].imageUrl.map((url, i) => (
+                  <Box
```
ksuhartono97, line 298: lol why are you using a whole box and setting the background image instead of just using an `Image` component?

ui/src/components/harvest-games/HarvestGameInstructions.js
```diff
@@ -0,0 +1,357 @@
...
+      ],
+      id: 'game_3',
+      pwd: 'Airplane',
+    },
+  ];
+
+  const [selected, setSelected] = useState(0);
+  const [password, setPassword] = useState('');
+
+  const handleSubmit = () => {
```
ksuhartono97, line 81: no error handling if password is wrong? i guess that works too but maybe some form of feedback to tell users that they have entered smth wrong? maybe `alert()`

ui/src/components/harvest-games/HarvestGameInstructions.js
```diff
@@ -0,0 +1,357 @@
...
+          </label>
+          <Input
+            id="inputted"
+            type="text"
+            placeholder="Start typing here"
+            w={{ base: '100%', sm: '50%' }}
+            backgroundColor="white"
+            borderWidth="1px"
+            borderColor="gray/300"
+            onKeyDown={(e) => handleKeyDown(e)}
```
ksuhartono97, line 341: side note, if you make this a form u don't need to deal with this but good workaround haha

ui/src/components/harvest-games/HarvestGamesPage.js
```diff
@@ -276,7 +289,7 @@ const HarvestGames = (props) => {
                 >
                   Leaderboard
                 </Tab>
-                <Tab
+                {/* <Tab
```
ksuhartono97, line 292: can you just delete the Pre-Games and Leaderboard sections and make sure just "Leaderboard" (which contains the Live Leaderboard) and "Game Instructions" are shown in both web and mobile?

ui/src/components/harvest-games/HarvestGamesPage.js
```diff
@@ -401,10 +425,13 @@ const HarvestGames = (props) => {
                     </Link>
                     .
                   </Text>
-                </TabPanel>
-                <TabPanel height={{ base: '80vh', sm: '90vh' }}>
+                </TabPanel> */}
+                <TabPanel minHeight={{ base: '30em', sm: '40em' }}>
```
ksuhartono97, line 429: not sure why you changed this since this actually breaks the harvest games leaderboard

ui/src/components/harvest-games/HarvestGamesPage.js
```diff
@@ -19,6 +19,7 @@ import {
 } from '@chakra-ui/react';
```
ksuhartono97, line 19: overall something happened with how u edited this file which is causing the `HarvestGamesLeaderboard` to not be rendering in it's tab. Can you check why your changes are breaking it?

ksuhartono97: image

ksuhartono97 (changes requested):

ui/src/components/harvest-games/HarvestGameInstructions.js
Baeck016: When I replace the image on the wordpress, the link still remains the same. So I think we can change even after we deploy it. So it isn't really hard coded.

ui/src/components/harvest-games/HarvestGameInstructions.js
Baeck016: It was easier to control the height haha.

ksuhartono97 (approved):

PR #581, opened 2022-10-14
GH-536: Add COVID 19 Policy Link in footer

andreajeka:


ui/src/components/footer/LinkGrid.js
```diff
@@ -72,6 +72,9 @@ export function LinkGrid() {
           <Link href="https://hongkong.sub.hmcc.net/privacy-policy-2/">
             <Text fontWeight="bold">Privacy Policy</Text>
           </Link>
+          <Link href="https://hongkong.sub.hmcc.net/urgent-announcements/hmcc-covid-19-safety-precautions/">
```
albertparedandan, line 75: Screenshot 2022-10-15 at 16 18 29

can we also add the same thing here?

albertparedandan (changes requested):

ui/src/components/footer/LinkGrid.js
andreajeka: okay good point

albertparedandan (approved):

PR #641, opened 2022-11-15
624 timer feature

cetair:


ui/src/components/advent/AdventCalendar.js
```diff
@@ -0,0 +1,74 @@
+import { Box, Grid, GridItem, Image } from '@chakra-ui/react';
+import AdventCard from './AdventCard';
+
+const AdventContainer = () => {
```
bhryck, line 4: i just realized from kris' commit that this is called "adventcontainer" even though it should be "adventcalender" since "adventcontainer" is another js file.

please help change this

PR #643, opened 2022-11-15
GH-636: mailing list form

bhryck:
expected behavior:
- similar to a popup where it will show up when the home page is rendered. 
- will not show up if there is a popup that is published to  avoid conflict
- hence we can just create a popup to remove the signup form for mailing list
- text and image placeholders are there but can be added

### DIFF
--- a/ui/src/components/home/MailingListForm.js
+++ b/ui/src/components/home/MailingListForm.js
@@ -56,10 +56,11 @@ const MailingListForm = ({ props }) => {
         category: 'advent',
       });
       toast({
-        description: 'Saved',
+        description: 'You have been added to the mailing list!',
         status: 'success',
         duration: 5000,
       });
+      setIsOpen(false);
     } catch (err) {
       console.log('Submitting to mailing list failed');
       setError('Submitting to mailing list failed');
@@ -149,7 +150,7 @@ const MailingListForm = ({ props }) => {
                   }}
                   colorScheme="teal"
                 >
-                  submit
+                  Submit
                 </Button>
               </ButtonGroup>
             </VStack>
--- a/ui/src/components/home/MailingListForm.js
+++ b/ui/src/components/home/MailingListForm.js
@@ -1,0 +1,163 @@
+import {
+  Box,
+  Button,
+  Link,
+  Modal,
+  ModalCloseButton,
+  Image,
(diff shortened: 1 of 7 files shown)


ui/src/components/home/MailingListForm.js
```diff
@@ -0,0 +1,163 @@
...
+  ButtonGroup,
+  useToast,
+} from '@chakra-ui/react';
+import { useState, useEffect } from 'react';
+import { useForm } from 'react-hook-form';
+import { customAxios as axios } from '../helpers/customAxios';
+
+const MailingListForm = ({ props }) => {
+  const user = props;
+  const [isOpen, setIsOpen] = useState(true);
```
ksuhartono97, line 26: can we make the default state check if the user has ever submitted to this haha, was noticing that this modal will always popup no matter what if there's no other modal present. thinking a better UX would be they just need to submit once and then it will disappear. also prevents multi submission

ui/src/components/home/MailingListForm.js
```diff
@@ -0,0 +1,163 @@
...
+                  maxW={['30em', '11em']}
+                  flex={[false, 1]}
+                  type="submit"
+                  style={{
+                    whiteSpace: 'normal',
+                    wordWrap: 'break-word',
+                  }}
+                  colorScheme="teal"
+                >
+                  submit
```
ksuhartono97, line 152: make it 'Submit' instead of submit

ui/src/components/home/MailingListForm.js
```diff
@@ -0,0 +1,163 @@
...
+  } = useForm();
+  const [error, setError] = useState('');
+
+  const handleSignup = async (data) => {
+    try {
+      const payload = await axios.post('/api/mailingList/create', {
+        email: data.email,
+        category: 'advent',
+      });
+      toast({
```
ksuhartono97, line 58: other than showing the toast, should also set the modal to just close so people are not left needing to close the modal

ui/src/components/home/MailingListForm.js
bhryck: added, also added a date check to only show when date is before dec 1

ksuhartono97 (approved):

PR #670, opened 2022-11-25
669 advent popup

bhryck:
- added popup image & additional field fullname
- added popup to advent page


ksuhartono97 (approved):
pls compress the adventad.png, if possible convert to jpg and compress to make it smaller size

PR #714, opened 2023-02-13
GH-712: easter apis

cetair:


server/api/controllers/easter/add-prayers.js
```diff
@@ -0,0 +1,46 @@
...
+    },
+    email: {
+      type: 'string',
+      required: false,
+    },
+    prayer: {
+      type: 'string',
+      required: true,
+    },
+    topics: {
```
alvone01, line 19: Minor feedback, I think it should be topic without the s (to avoid confusion)

server/api/controllers/easter/get-prayers.js
```diff
@@ -0,0 +1,48 @@
+module.exports = {
+  friendlyName: 'Get prayers',
+
+  description: 'Get all prayers',
+
+  inputs: {
+    prayer: {
```
alvone01, line 7: curious, why is prayer part of the input?

PR #716, opened 2023-02-13
change time format

Baeck016:
Change time format.


ksuhartono97:
addressed with different PR

PR #739, opened 2023-03-12
GH-732: Fix User membership recognition bug

albertparedandan:


ui/src/components/admin/users/AdminUser.js
```diff
@@ -200,21 +173,18 @@ export default function AdminUser(props) {
...
+  const membershipBaptismDateFormatter = (params) => {
+    if (params.value) {
+      const dateTimeFormat = 'dd MMM yyyy';
+      const dateTimeObj = DateTime.fromISO(params.value);
+      if (dateTimeObj.isValid)
+        return dateTimeObj.toFormat(dateTimeFormat);
     }
 
-    return '';
+    return ''
```
tamarayustian, line 184: minor, just missing semi-colon. no big deal

ui/src/components/admin/users/AdminUser.js
```diff
@@ -444,9 +414,12 @@ export default function AdminUser(props) {
       ...data,
     });
 
-    if (res.status !== 200) {
+    if (res.status !== 200 || res.data.id == null || res.data.id === '') {
```
tamarayustian, line 417: why for the comparison of res.data.id, you used 2 different equal signs: '==' and '==='?  just curious. i dont think it affects anything

server/api/controllers/memberships/update-membership.js
```diff
@@ -26,10 +26,26 @@ module.exports = {
...
-    const { id: membershipId, ...toUpdate } = params;
+    
+    const { id: membershipId, recommitmentDate, ...toUpdate } = params;
     if (membershipId) {
       try {
+        if (recommitmentDate && recommitmentDate !== '') {
+          let data = await Membership.findOne({ _id: membershipId });
+
+          if (data == null)
+            return exits.invalid(`No MembershipId found id: ${membershipId}`);
```
tamarayustian, line 37: error msg can just be 'No MembershipId found' to be grammatically correct

tamarayustian (approved):

ui/src/components/admin/users/AdminUser.js
albertparedandan: `== null` checks if `id` is `null` or `undefined`, `=== '"` checks if `id` is string type and is not empty string

PR #772, opened 2023-06-26
GH-769: Enable External Forms

ksuhartono97:
Enable externally linked forms

### DIFF
--- a/server/api/controllers/forms/get-form.js
+++ b/server/api/controllers/forms/get-form.js
@@ -15,9 +15,7 @@ module.exports = {
 
   fn: async function ({ id }, exits) {
     try {
-      let now = new Date();
-      let offset = now.getTimezoneOffset();
-      now.setTime(now.getTime() - offset * 60000);
+      let now = await sails.helpers.datetime.getOffsettedTime(new Date());
       now = now.toISOString();
 
       const data = await Form.find({
--- a/server/api/controllers/forms/get-form.js
+++ b/server/api/controllers/forms/get-form.js
@@ -15,10 +15,26 @@ module.exports = {
 
   fn: async function ({ id }, exits) {
     try {
+      let now = new Date();
+      let offset = now.getTimezoneOffset();
+      now.setTime(now.getTime() - offset * 60000);
+      now = now.toISOString();
+
       const data = await Form.find({
         _id: id,
         isPublished: true,
         isDeleted: false,
+        or: [
+          {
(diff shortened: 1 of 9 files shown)


ui/src/components/forms/ExternalFormEditor.js
```diff
@@ -0,0 +1,142 @@
...
+              <Input
+                id="externalFormLink"
+                {...register('externalFormLink', { required: true })}
+              />
+              <FormErrorMessage>
+                {errors['externalFormLink'] && 'External Form Link is required'}
+              </FormErrorMessage>
+            </FormControl>
+            <Button mt={4} colorScheme="blue" type="submit">
+              Save External Form Data
```
bhryck, line 133: when clicking external form 'view form' in admin page it doesn't work
but it does work on my profile page. wondering if this is intended behaviour

server/api/controllers/forms/get-form.js
```diff
@@ -15,10 +15,26 @@ module.exports = {
 
   fn: async function ({ id }, exits) {
     try {
+      let now = new Date();
+      let offset = now.getTimezoneOffset();
+      now.setTime(now.getTime() - offset * 60000);
+      now = now.toISOString();
+
```
bhryck, line 22: imo we can move this to a separate helper function, may need to use it again in the future

ui/src/components/forms/FormManager.js
```diff
@@ -150,20 +147,26 @@ const FormManager = (props) => {
...
             {formList.map((formItem) => (
-              <ListItem key={formItem.id}>
+              <Box key={formItem.id}>
                 <Box p="3" borderRadius="lg" borderWidth="1px">
                   <Text mb="3">
                     <Badge colorScheme={formItem.isPublished ? 'green' : 'red'}>
                       {formItem.isPublished ? 'LIVE' : 'PRIVATE'}
                     </Badge>{' '}
+                    <Badge>
+                      {formItem.formType ? formItem.formType : 'internal'}
```
bhryck, line 162: when clicking external form 'view form' in admin page it doesn't work
but it does work on my profile page. wondering if this is intended behaviour

ui/src/components/forms/ExternalFormEditor.js
bhryck, line 133: would prefer 'save form link' here to clarify we're only saving one field

bhryck (approved):

PR #779, opened 2023-07-03
GH-778:Automate Live Button to turn on and off

JoshNg123:


server/api/controllers/liveSermon/scheduler-live-sermon.js
```diff
@@ -0,0 +1,60 @@
+const cron  = require('node-cron');
+
+// To change the isPublished value in the database from false to true
+const onLiveButon = async () => {
+  try{
+    var mongoClient = LiveSermon.getDatastore().manager.client;
+    var results = await mongoClient.db('hmcchk_db_test')
```
bhryck, line 7: Correct me if im wrong, but this seems to only work with the test db. it is better to use environment variables instead of calling the DB name directly. 

@ksuhartono97 @albertparedandan  please advice what the env variable name is in the prod server

server/api/controllers/liveSermon/scheduler-live-sermon.js
```diff
@@ -0,0 +1,60 @@
...
+        { $set: { isPublished: true } },
+      )
+    }
+  }catch(err){
+    console.log(err);
+  }
+}
+
+// To change the isPublished value in the database from true to false
+const offLiveButton = async () => {
```
bhryck, line 27: onLiveButton and offLiveButton has a lot of code overlap, could we make change it to a function swtichLiveButton(off/on) instead

server/api/controllers/liveSermon/scheduler-live-sermon.js
```diff
@@ -0,0 +1,60 @@
+const cron  = require('node-cron');
```
albertparedandan, line 1: have you tested if this whole logic works? i don't see anywhere where we're calling the cron job to start

albertparedandan: also, quite a loaded comment, but i don't think this should be the way we implement this. we already have a cron job running for caching (take a look at `server/config/bootstrap.js`). We should just be extending that

albertparedandan: lmk if you want clarfication

albertparedandan: actually, this is making it way too complicated. why do we need a cron job? why can't we just set LiveSermon object to have a `StartTime` and `EndTime` that will up and down according to the times?

albertparedandan (changes requested):

server/api/models/LiveSermon.js
```diff
@@ -38,5 +38,13 @@ module.exports = {
     lastUpdatedBy: {
       type: 'string',
     },
+    streamStartTime: {
```
albertparedandan, line 41: please remove `isPublished` from the model

server/api/controllers/liveSermon/get-live-sermon.js
```diff
@@ -11,7 +13,7 @@ module.exports = {
     isPublished: {
```
albertparedandan, line 13: remove too

server/api/controllers/liveSermon/get-live-sermon.js
```diff
@@ -28,16 +30,36 @@ module.exports = {
...
         }).populateAll();
         if (data.length === 0) {
           return exits.success([]);
         }
         return exits.success(data);
       }
 
-      let data = await LiveSermon.find({ isDeleted: false, isPublished }).sort('updatedAt DESC').populateAll();
-      sails.log.info('Retrieving live sermons');
+      let data = await LiveSermon.find({ isDeleted: false, isPublished })
```
albertparedandan, line 41: i would make a couple changes here.
1. remove `isPublished` as API's input params
2. If a `sermonId` passed in input, simply find and return.
3. If `sermonId` not passed as input, find the most recently updated sermon that has `startTime < endTime && now > startTime && now < endTime`

albertparedandan: you will have to also search up the whole codebase for pages that calls this API and update them accordingly

ui/src/components/admin/liveSermon/AdminLiveSermonContainer.js
```diff
@@ -48,10 +51,11 @@ export default function AdminLiveSermonContainer(props) {
     setSermonSeriesUrl(data.sermonSeriesUrl);
     setSermonPassage(data.sermonPassage);
     setSermonDateTime(data.sermonDateTime);
-    setIsPublished(data.isPublished);
+    setStreamStartTime(data.streamStartTime);
+    setStreamEndTime(data.streamEndTime);
   }
 
-  const sanityCheckFailed = title === '' || sermonDescription === '';
+  const sanityCheckFailed = title === '' || sermonDescription === '' || streamPeriodInvalid === true;
```
albertparedandan, line 58: can simplify to `=== streamPeriodInvalid`

ui/src/components/admin/liveSermon/AdminLiveSermonContainer.js
```diff
@@ -194,6 +203,18 @@ export default function AdminLiveSermonContainer(props) {
     getData();
   }, [getData]);
 
+  useEffect(() => {
+    if (streamStartTime && streamEndTime) {
+      const startTime = DateTime.fromISO(streamStartTime);
+      const endTime = DateTime.fromISO(streamEndTime);
+      if (startTime >= endTime) {
+        setStreamPeriodInvalid(true);
+        return;
```
albertparedandan, line 212: no need to `return` here

ui/src/components/admin/liveSermon/AdminLiveSermonContainer.js
```diff
@@ -194,6 +203,18 @@ export default function AdminLiveSermonContainer(props) {
...
+      }
+    }
+    setStreamPeriodInvalid(false);
```
albertparedandan, line 215: also no need to `setStreamPeriodInvalid` to `false` since you've initialised it as `false` anyway

albertparedandan (changes requested):

ui/src/components/admin/liveSermon/AdminLiveSermonContainer.js
JoshNg123: For this change and the change 'no need to return here' have to be put because it will always be set to true and I cannot update the sermon

ksuhartono97 (approved):

PR #780, opened 2023-07-03
Add more fields to announcements model and update announcements API

tamarayustian:


server/config/routes.js
```diff
@@ -115,6 +115,7 @@ module.exports.routes = {
   'get /api/announcement/get': 'announcements/get-announcements-new',
   'get /api/announcement/admin-get': 'announcements/admin-get-announcements',
   'post /api/announcement/create': 'announcements/create-announcements',
+  'put /api/announcement/update': 'announcements/update-announcements',
```
bhryck, line 118: you will have to modify routes.js for the new APIs
` eg. in routes js:
  'announcements/*': true,`

tamarayustian: hmm you right. can be a fix for future

bhryck (approved):

PR #790, opened 2023-07-19
GH-759: Hide dates, locations if undefined, change api to new one

Midnite1212:


ui/src/components/helpers/eventsHelpers.js
```diff
@@ -13,11 +13,17 @@ const getStartDate = (eventData) => {
 };
 
 const getRenderDate = (startDate, endDate, interval) => {
+  // parse the interval to number
+  let recur = interval === "Daily" ? 1 : 
+              interval === "Weekly" ? 7 : 
+              interval === "Monthly" ? 30 :
```
bhryck, line 19: idt we can assume a month is always 30 days

bhryck: you can modify the snippet at the bottom to use months instead of offset by days  `  for (let i = 0; i <= nRecurrence; i++) {
    renderDate = start.plus({ days: interval * i });
    if (DateTime.now() < renderDate) break;
  }`

ksuhartono97 (approved):

PR #798, opened 2023-08-05
GH-1: add tc and mt access rights to pre-existing pages

tamarayustian:


ui/src/components/MainContainer.js
```diff
@@ -127,6 +127,8 @@ const MainContainer = () => {
             't3ch',
             'admin',
             'stewardship',
+            'tc',
```
albertparedandan, line 130: What's the diff between this and #114

ui/src/components/MainContainer.js
```diff
@@ -202,6 +208,8 @@ const MainContainer = () => {
...
+            'ministry',
```
albertparedandan, line 212: /profile and /complete-profile is duplicated

ui/src/components/MainContainer.js
tamarayustian: hm i dont get it? this is just to allow TC and MT users to still allow them use their profiles as normal. because I tested with someone in TC and they can access `/admin/forms` but not `/profile`

albertparedandan: My bad, didn't actually comment on the offending line. Both #114 and #122 have the same link but different permissions which doesn't make sense to me

albertparedandan (changes requested):
Commented

ui/src/components/MainContainer.js
tamarayustian: oooo ok hmm. cmiiw but is this because some forms requires login and some are not? @ksuhartono97 
image

ksuhartono97 (approved):

albertparedandan (approved):

PR #806, opened 2023-09-10
GH-1: Change LG name

tamarayustian:


albertparedandan (approved):

albertparedandan:
there's got to be a better way of doing this..

PR #827, opened 2023-10-12
GH-826: Fundraise APIs

tamarayustian:


server/api/models/Fundraise.js
```diff
@@ -0,0 +1,37 @@
...
+  attributes: {
+    campaignName: {
+      type: 'string',
+      required: true,
+    },
+    categoryName: {
+      type: 'string',
+      required: true,
+    },
+    categoryKey: {
```
bhryck, line 11: whats category key and name for?

server/api/models/Fundraise.js
```diff
@@ -0,0 +1,37 @@
...
+      type: 'string',
+      required: true,
+    },
+    amount: {
+      type: 'number',
+    },
+    givers: {
+      type: 'number',
+    },
+    milestones: {
```
bhryck, line 21: im assuming the usecase will be something like:
[
  {
    "value": "0",
    "description": "start"
  },
  {
    "value": "50",
    "description": "we have enough money to buy a car"
  },
  {
    "value": "100",
    "description": "fundraising goal"
  }
]

tamarayustian: yes, i added more comments and examples now to make it clear in the code

server/api/models/Fundraise.js
tamarayustian: in giving tuesday, there are categories of giving so we want to account for that. the example would be
```
categoryName: Food for the poor
categoryKey: food-for-the-poor
```

bhryck (approved):

server/api/controllers/fundraise/create-fundraise.js
```diff
@@ -0,0 +1,60 @@
...
+    },
+    amount: {
+      type: 'number',
+    },
+    givers: {
+      type: 'number',
+    },
+    milestones: {
+      type: 'json',
+    },
```
bhryck, line 27: had an error when testing api out 
`error: Failed to lift app: userError: The action `fundraise/create-fundraise` could not be registered.  It looks like a machine definition (actions2), but it could not be used to build an action.
Details: ImplementationError: Sorry, could not interpret "fundraise/create-fundraise.js" because its underlying implementation has 4 problems:
------------------------------------------------------
• Invalid input name ("exits").  Please use something else instead, because this is potentially very confusing.
• Invalid input definition ("exits").  Must have `type`, or at least some more information.  (If you aren't sure what to use, just go with `type: 'ref'.)
• Invalid input definition ("fn").  Must be a dictionary-- i.e. plain JavaScript object like `{}`.
• Invalid input definition ("fn").  Must have `type`, or at least some more information.  (If you aren't sure what to use, just go with `type: 'ref'.)`

I think you missed an exit '}' here

ksuhartono97 (approved):

PR #833, opened 2023-10-12
GH-816: Admin FAQ Management Page

yuenyauu:
- Added an admin management page for FAQs

### DIFF
--- a/ui/src/components/MainContainer.js
+++ b/ui/src/components/MainContainer.js
@@ -41,6 +41,7 @@ import AdminAnnouncementContainer from './admin/announcements/AdminAnnouncementC
 import PrayerFloatButton from './connect-prayer/PrayerFLoatButton';
 import RippleOutContainer from './ripple-out/RippleOutContainer';
 import AdminFundraiseContainer from './admin/fundraise/AdminFundraiseContainer';
+import AdminFaqContainer from './admin/faqs/AdminFaqContainer';
 
 const MainContainer = () => {
   return (
@@ -244,6 +245,12 @@ const MainContainer = () => {
           path="/admin/announcements"
           permissions={['ministry', 'tc', 't3ch', 'admin', 'stewardship']}
           component={AdminAnnouncementContainer}
+        />
+        <PrivateRoute
+          exact
+          path="/admin/faqs"
+          permissions={['t3ch', 'admin', 'stewardship']}
+          component={AdminFaqContainer}
         />
         <PrivateRoute
           exact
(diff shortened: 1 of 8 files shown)


ui/src/components/MainContainer.js
```diff
@@ -244,6 +245,12 @@ const MainContainer = () => {
           permissions={['ministry', 'tc', 't3ch', 'admin', 'stewardship']}
           component={AdminAnnouncementContainer}
         />
+        <PrivateRoute
+          exact
+          path="/admin/faqs"
+          permissions={['t3ch', 'admin', 'stewardship']}
```
tamarayustian, line 251: should we give this access to tc too? in case they are submitting FAQ for their crew? @ksuhartono97 
which makes me think we should probably make a guide for this if that's the case l o l

ui/src/components/admin/faqs/FaqGrid.js
```diff
@@ -0,0 +1,90 @@
...
+    if (faqs) {
+      if (!showDeleted) {
+        setFiltered(faqs.filter((p) => p.isDeleted === false));
+      } else {
+        setFiltered(faqs);
+      }
+    }
+  }, [faqs, showDeleted]);
+
+  useEffect(() => {
```
tamarayustian, line 25: can you add comments to these 2 useEffects? help improve readability for future

ui/src/components/admin/faqs/FaqGrid.js
```diff
@@ -0,0 +1,90 @@
+import React, { useState, useEffect } from 'react';
+import { AgGridReact } from 'ag-grid-react';
+import 'ag-grid-community/dist/styles/ag-grid.css';
+import 'ag-grid-community/dist/styles/ag-theme-alpine.css';
+import { Heading, Container, Switch, HStack } from '@chakra-ui/react';
+
+export default function FaqGrid(props) {
+  const { faqs, setSelected } = props;
```
tamarayustian, line 8: can combine this to be `export default function FaqGrid({ faqs, setSelected }) {...`

yuenyauu: i think the convention is to use props hehe, so maybe we can keep it as props instead?

tamarayustian: can too idm. i just like shorter code hahha

tamarayustian (approved):
might need to give access to TC in FAQ page if needed in the future

tamarayustian (approved):

ksuhartono97 (approved):

PR #835, opened 2023-10-14
GH-826: Created funds progress bar

alvone01:


ui/src/components/ripple-out/ProgressBar.js
```diff
@@ -0,0 +1,90 @@
+import Tippy from '@tippyjs/react';
```
bhryck, line 1: what does tippy do exactly

ksuhartono97 (approved):

PR #846, opened 2023-10-15
GH-844: Add pledge section and fix extra padding bug

tamarayustian:


ui/src/components/ripple-out/ProgressSection.js
```diff
@@ -27,8 +27,9 @@ import MilestoneProgressBar from './MilestoneProgressBar';
 //   ]
 // }
 
-const ProgressSection = () => {
+const ProgressSection = ({ bodyFontSize }) => {
   const [fundraiseData, setFundraiseData] = useState(null);
+  const [pledeData, setPledgeData] = useState(null);
```
ksuhartono97, line 32: typo here, `pledgeData` instead of "plede" haha so need to refactor for the rest too

ui/src/components/admin/fundraise/AdminFundraiseContainer.js
```diff
@@ -120,8 +117,7 @@ export default function FundraiseContainer(props) {
   };
   const onSubmit = async (event) => {
     event.preventDefault();
-    if (!campaignName || campaignName.length === 0 || campaignNameCheck())
```
ksuhartono97, line 123: what was this for actually lol not sure why it's removed

tamarayustian: 1 campaign can have multiple categories, so it needs more than 1 object. so campaignName has to be duplicated. if u look at the backend, i actually just create another category called 'pledge'

PR #855, opened 2023-10-21
GH-816: FAQ Reordering and Sync to Public Page

yuenyauu:
- Added order property to FAQ model
- Added FAQ reordering feature in admin panel
- Synced public page with FAQ database


server/api/controllers/faq/create-faq.js
```diff
@@ -24,14 +28,15 @@ module.exports = {
 
   exits: {},
 
-  fn: async function ({ pageTopic, question, answer, isPublished }, exits) {
+  fn: async function ({ pageTopic, question, answer, order, isPublished }, exits) {
     const createdBy = this.req.user.fullName;
 
     try {
       const newFaq = await Faq.create({
```
bhryck, line 35: image

thinking we could use a more graceful error msg

bhryck: or maybe even set the page topic field as a drop down

tamarayustian: oh yea since it's already enum, can be a dropdown list. just put it in /lists.js

yuenyauu: ok!

ui/src/components/admin/faqs/FaqGrid.js
```diff
@@ -2,26 +2,59 @@ import React, { useState, useEffect } from 'react';
...
+  const { faqs, setSelected, toast, resetHandler, updateHandler, isLoading } =
+    props;
 
+  const [pageTopicFilter, setPageTopicFilter] = useState('');
   const [showDeleted, setShowDeleted] = useState(false);
+  const [showPublishedOnly, setShowPublishedOnly] = useState(false);
   const [filtered, setFiltered] = useState(faqs);
+  const [isReorderingFaqs, setIsReorderingFaqs] = useState(false);
+  const [toPublish, setToPublish] = useState(false);
+  const [toUnpublish, setToUnpublish] = useState(false);
```
tamarayustian, line 27: Why need both `toPublish`, `toUnpublish`? don't they work the same?

ui/src/components/admin/faqs/FaqGrid.js
```diff
@@ -34,24 +67,201 @@ export default function FaqGrid(props) {
...
+      }
+    }
+  };
+  // Enables FAQ reordering
+  const enableFaqReordering = () => {
+    setShowDeleted(false);
+    setIsReorderingFaqs(true);
+  };
+  // Cancels FAQ publishing/reordering
+  const cancelHandler = () => {
```
tamarayustian, line 183: oh i get it now why.
can you just do this? right now you're forcing the behaviour and it kinda looks like hard coding to me. same applies for line 152
```
  const cancelHandler = () => {
    toPublishList.forEach((faq) => {
      faq.setDataValue('isPublished', !toPublish);
    });
    setToPublish(!toPublish);
    setIsReorderingFaqs(false);
  };
```

ui/src/components/admin/faqs/FaqGrid.js
```diff
@@ -71,10 +281,68 @@ export default function FaqGrid(props) {
         <Switch
```
tamarayustian, line 281: i just realixed the <Heading> here has `s` and `size`. if s is not size, what is it?

ui/src/components/admin/faqs/FaqGrid.js
```diff
@@ -34,24 +67,201 @@ export default function FaqGrid(props) {
...
+      });
+    }
+  };
+
+  // FAQ Update Functions
+  const saveFaqInfo = async (data) => {
+    return updateHandler(data);
+  };
+
+  const saveFaqReordering = async () => {
```
tamarayustian, line 211: might be better to be in a try catch. right now if there is no `api`, you'll never know if it runs

tamarayustian (changes requested):

ui/src/components/admin/faqs/FaqGrid.js
yuenyauu: haha i dont know, ill remove it

ui/src/components/admin/faqs/AdminFaqContainer.js
```diff
@@ -94,7 +94,7 @@ export default function AdminFaqContainer(props) {
         pageTopic: pageTopic,
         question: question,
         answer: answer,
-        isPublished: published,
+        order: order === '' ? -1 : order,
```
ksuhartono97, line 97: realized this has no duplicate order checking so we might have duplicate orders. not sure how necessary to fix since things look like it's working ok

image

ui/src/components/admin/faqs/FaqGrid.js
```diff
@@ -34,24 +86,217 @@ export default function FaqGrid(props) {
...
+      toast({
+        description: 'Successfully reordered FAQs',
+        status: 'success',
+        duration: 5000,
+      });
+
+      // Refresh FAQs with updated data
+      await getData();
+
+      setIsReorderingFaqs(false);
```
ksuhartono97, line 258: copy these setters which i think is for initializing down to the catch case too. otherwise the frontend might be stuck when there's an error

ui/src/components/ripple-out/RippleOutContainer.js
ksuhartono97, line 1: this whole file shouldn't exist anymore with richie's new PR

ui/src/components/ripple-out/RippleOutContainer.js
```diff
@@ -0,0 +1,345 @@
...
+      </Flex>
+
+      {/* FAQ section */}
+      <Flex
+        flexDir="column"
+        bgColor="#C9DDED"
+        py={['1.8em', '4em']}
+        color={fontColor}
+      >
+        <RippleOutFaqSection headerFontSize={headerFontSize} />
```
ksuhartono97, line 339: move this to the appropriate page according to the design. right now the FAQ appears on every page

ksuhartono97 (changes requested):

ui/src/components/admin/faqs/AdminFaqContainer.js
yuenyauu: realized i didn't add a check for the order update through the grid (i.e. if FAQ is being unpublished, set order to -1), i think should work fine after i add it

ui/src/components/admin/faqs/FaqGrid.js
yuenyauu: ok!

ksuhartono97 (approved):

tamarayustian (approved):

Issue #852, opened 2023-10-21
Hero section and our story


bhryck:

bhryck:
![Image](https://github.com/hmcc-global/hmcchk-web/assets/24607458/168c9907-dc03-494a-b91c-f61e31d5472d)

PR #869, opened 2023-10-28
GH-852: Ripple Out Story Revamp

yuenyauu:
- Added Mission Circles in the form of a .png
- Added Story Carousel
- Updated @chakra-ui/react package


ui/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@chakra-ui/icons": "^1.0.15",
-    "@chakra-ui/react": "^1.6.5",
```
ksuhartono97, line 7: hmm will this be a breaking change for the rest of our site... not sure if this might break some of the things that our site uses.

can you verify that none of the other functionalities in our site break because of this lol

ui/src/components/ripple-out/RippleOutStoryContainer.js
```diff
@@ -33,111 +108,62 @@ const RippleOutStoryContainer = () => {
...
-              fontSize="lg"
-              bgColor="#ffffff"
-              fontWeight="bold"
-              borderColor="#182E57"
-              color="#182E57"
+                We are called to ripple out as we live out our Saturate vision
+                to saturate Hong Kong with the knowledge of God’s glory
+              </Highlight>
+            </Text>
+            <Image
```
ksuhartono97, line 133: image

so the sides of the image gets affected by the padding from the container. i think u might need to take the image out of the container so that it can break the margin rules of the container cos according to the design this is edge to edge on the screen

ui/src/components/ripple-out/RippleOutStoryContainer.js
```diff
@@ -33,111 +108,62 @@ const RippleOutStoryContainer = () => {
...
-            </Button>
+              TRANSFORMING LIVES. TRANSFORMING THE WORLD.
+            </Text>
+            <Text
+              fontSize={['1.65em', '1.80em', '2.40em', '2.65em', '2.9em']}
+              fontWeight={600}
+            >
+              SWIPE TO READ OUR STORY {'>>'}
+            </Text>
+            <Slider {...sliderSettings} style={sliderStyle}>
```
ksuhartono97, line 155: @Midnite1212 can steal from this for swipe behavior

ksuhartono97: but also, missing left/right arrows from the design?

image

ksuhartono97 (changes requested):

ui/src/components/ripple-out/RippleOutStoryContainer.js
ksuhartono97: also can u confirm with CXO if they want this to be a gif, saw that on figma but not sure

ui/src/components/ripple-out/RippleOutStoryContainer.js
yuenyauu: eh the arrows should only disappear on mobile and iPad mini view... did you start off with desktop view?

ui/src/components/ripple-out/RippleOutStoryContainer.js
yuenyauu: no gif!

ui/package.json
yuenyauu: from what i've tested, nothing has been broken...

ui/src/components/ripple-out/RippleOutStoryContainer.js
ksuhartono97: oh so the arrows aren't reactive? like if i switch over from desktop to mobile to desktop viewports it affects the arrow visibillity?

ksuhartono97 (approved):

Issue #884, opened 2023-11-02
ROC The Space Fixes


melissaprayogo:
make it the same size as the 'crocodile centre' circle.
![IMG_4995](https://github.com/hmcc-global/hmcchk-web/assets/100053797/173e9360-9272-4bae-a439-44fca98fcd5e)

yuenyauu:
![image](https://github.com/hmcc-global/hmcchk-web/assets/66002742/a96c0727-f91c-4527-b620-b77175866e64)
- make Crocodile Centre larger

PR #896, opened 2023-11-04
GH-884: ROC The Space Fixes

yuenyauu:
- Increased tooltip sizes
- Increased font size for Crocodile Centre
- Adjusted tooltip positioning

### DIFF
--- a/ui/src/components/ripple-out/RippleOutSpaceContainer.js
+++ b/ui/src/components/ripple-out/RippleOutSpaceContainer.js
@@ -85,7 +85,7 @@ const RippleOutSpaceContainer = () => {
                     w="6em"
                     h="3em"
                     textStyle="darker_grotesque_semibold"
-                    fontSize="lg"
+                    fontSize={['1.1em', 'xl']}
                   >
                     Crocodile Center
                   </Box>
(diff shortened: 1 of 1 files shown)


tamarayustian (approved):
interesting that you do negative. why not use bottom (like opposite direction) instead of negative with top? approved anyway tho

ui/src/components/ripple-out/RippleOutSpaceContainer.js
```diff
@@ -149,13 +149,13 @@ const RippleOutSpaceContainer = () => {
               placement="top"
             >
               <Box
-                top={['-6vh', '-14vh']}
-                left={['-30vw', '-18vw']}
+                top={['-6.5vh', '-17vh']}
```
ksuhartono97, line 152: image
the dots now look misaligned to me? the building blocks dot is now out of the room and the office dot is covering the text

ksuhartono97 (changes requested):

ui/src/components/ripple-out/RippleOutSpaceContainer.js
yuenyauu: image
looks good on mine though lolol

yuenyauu: i guess i could make the buttons slightly smaller

tamarayustian (commented):
image
since you're on this file, can you also remove these unnecessary imports? thanks!

ksuhartono97 (approved):

Issue #906, opened 2023-11-10
about page nav links should be 1 line but appearing as 2 lines (the our story, vision mission, ...)


Midnite1212:
<img width="653" alt="image" src="https://github.com/hmcc-global/hmcchk-web/assets/56828805/49579659-6c30-4fd5-9afb-598708e65649">

PR #910, opened 2023-11-10
GH-906: nav links 1 line

cetair:


albertparedandan (approved):

albertparedandan (changes requested):
this needs to be pointed to main instead of ripple-out

albertparedandan (approved):

Issue #905, opened 2023-11-10
Connect page, buttons under "experience hmcc" color is broken and not legible


Midnite1212:
<img width="615" alt="image" src="https://github.com/hmcc-global/hmcchk-web/assets/56828805/a5ae02d8-061a-4fe1-aae0-2cb7827d9137">

PR #909, opened 2023-11-10
GH-905: inherit color

cetair:

PR #914, opened 2023-11-12
GH-905: inherit color

ksuhartono97:
recover from original branch


ksuhartono97:
originally authored by @cetair here https://github.com/hmcc-global/hmcchk-web/pull/909

Issue #908, opened 2023-11-10
Events page sorting


Midnite1212:
new feature request, can we adjust the logic of the announcements rendering (both here and the homepage carousel) so that events that don't have date like LG signup and BRP should be shown at the end of the list

PR #918, opened 2023-11-12
GH-908: replicate changes on homepage

ksuhartono97:


tamarayustian (approved):
emg cmn gini? bkannya ada buttons jg

PR #928, opened 2024-01-28
GH-1: hotfix announcements logic

Midnite1212:
Fix announcement logic, no date announcement will be shown in the back (Fixed logic to handle multiple no date announcements). Announcements will be sorted from newest dates.

### DIFF
--- a/ui/src/components/home/EventsSectionCards.js
+++ b/ui/src/components/home/EventsSectionCards.js
@@ -76,31 +76,24 @@ const EventsSectionCard = (props) => {
   };
 
   const getStartEndDateStr = (eventData) => {
+    const eventRenderDate = DateTime.fromISO(
+      getRenderDate(
+        eventData.eventStartDate,
+        eventData.eventEndDate,
+        eventData.eventInterval,
+        eventData.eventStartTime
+      )
+    );
+    const dateFormat = eventData.eventStartTime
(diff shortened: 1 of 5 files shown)


tamarayustian:
@Baeck016 should we make the interval field also be put as default Daily? i'm thinking since you said thats the fix you made for the events carousell at home page.
it doesnt make sense i think because sometimes there is no recurring for one-time events. but i dont know what is the code issue there

Midnite1212:
> @Baeck016 should we make the interval field also be put as default Daily? i'm thinking since you said thats the fix you made for the events carousell at home page.
> it doesnt make sense i think because sometimes there is no recurring for one-time events. but i dont know what is the code issue there

I fixed that issue alrd, now they can input the interval as none. Doesnt always have to be daily.

ui/src/components/home/EventsSectionCards.js
```diff
@@ -76,31 +76,24 @@ const EventsSectionCard = (props) => {
   };
 
   const getStartEndDateStr = (eventData) => {
+    const eventRenderDate = DateTime.fromISO(
```
tamarayustian, line 79: nice i like this. its cleaner looking code

tamarayustian (approved):

Midnite1212:
Closing this pr since its based off the wrong branch

Issue #936, opened 2024-02-18
Baptism Screen Workflow


ksuhartono97:
Screen containing alerts on members to follow up for baptism

PR #975, opened 2024-03-10
GH-936: Add BaptismFollowUp screen

albertparedandan:


ui/src/components/admin/follow-up/BaptismFollowUpContainer.js
```diff
@@ -0,0 +1,46 @@
+import React, { useEffect, useState } from 'react';
```
tamarayustian, line 1: no need to import `React`

ui/src/components/MainContainer.js
```diff
@@ -247,6 +248,12 @@ const MainContainer = () => {
           permissions={['admin', 'stewardship']}
           component={AdminUser}
         />
+        <PrivateRoute
+          exact
+          path="/admin/followUp"
```
tamarayustian, line 253: is it better if the path name is `/admin/followUp/baptism`? since this is specifically for baptism only

ui/src/components/admin/follow-up/UserGrid.js
```diff
@@ -0,0 +1,411 @@
+import React, { useState, useEffect } from 'react';
```
tamarayustian, line 1: no need `React`

ui/src/components/admin/follow-up/UserGrid.js
```diff
@@ -0,0 +1,411 @@
...
+};
+
+const tithelyIdGetter = (params) => {
+  const { tithelyId } = params.data;
+  if (!tithelyId || tithelyId === '') return '';
+
+  return tithelyId.join(',');
+};
+
+// Formatters
```
tamarayustian, line 151: actually all this formatters, getters, and setters is repeated from user table ya? should we have helpers file for functions and import it then? to avoid repeated code

ui/src/components/admin/follow-up/BaptismFollowUpContainer.js
albertparedandan: I might be wrong but all react component files need import React even tho not explicitly being used

ui/src/components/MainContainer.js
albertparedandan: Cab do once we add the other follow up pages

ui/src/components/admin/follow-up/UserGrid.js
albertparedandan: Yes. But that requires quite a bit of refactor.

tamarayustian: i guess can put //TODO for future then

ui/src/components/admin/follow-up/BaptismFollowUpContainer.js
tamarayustian: oh i never do, but ok

tamarayustian (approved):
just add the // TODO comment please for the refactor of that code for getters, setters, etc so we remember in future. otherwise LGTM woohoo

ksuhartono97:
@albertparedandan i adjusted the merge branch to release-database, not sure why you wanted to merge to main for this one ya

Issue #939, opened 2024-02-18
LG leaders dashboard


ksuhartono97:

tamarayustian:
For LG leaders frontend:
- [ ]  make the RESET button here more intuitive looking.
- [ ]  also make these buttons to be sticky so you can see it even when you scroll to the bottom of the table

PR #981, opened 2024-03-23
GH-939: LG leaders frontend

tamarayustian:
After https://github.com/hmcc-global/hmcchk-web/pull/979 is merged, then can merge this

### DIFF
--- a/ui/src/components/admin/leadershipTeam/AdminLeadershipTeamContainer.js
+++ b/ui/src/components/admin/leadershipTeam/AdminLeadershipTeamContainer.js
@@ -89,6 +89,7 @@ export default function AdminLeadershipTeamContainer(props) {
   const createHandler = async () => {
     try {
       const leadersArr = leaders && leaders.length > 0 && leaders.split(',');
+      const cleanLeadersArr = leadersArr.map((x) => x.trim());
 
       const res = await axios.post('/api/leadership-team/create', {
         seasonFrom,
@@ -96,38 +97,7 @@ export default function AdminLeadershipTeamContainer(props) {
         campus,
         lifestage,
         lifeGroup,
-        leaders: leadersArr,
-        isDeleted: deleted,
-      });
-
-      if (res.status === 200) {
-        const { data } = res;
-        setId(data.id);
-        return true;
-      }
-    } catch (e) {
-      console.log(e.response);
-      toast({
-        description: e.response.data,
-        status: 'error',
-        duration: 5000,
-      });
-      return false;
-    }
-  };
-
-  const updateHandler = async () => {
-    try {
-      const leadersArr = leaders && leaders.length > 0 && leaders.split(',');
-
-      const res = await axios.put('/api/leadership-team/update', {
-        id,
-        seasonFrom,
-        seasonTo,
-        campus,
-        lifestage,
-        lifeGroup,
-        leaders: leadersArr,
+        leaders: cleanLeadersArr,
         isDeleted: deleted,
       });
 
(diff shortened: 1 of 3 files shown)


ksuhartono97 (commented):
code content actually includes both FE and BE changes? duplicate of https://github.com/hmcc-global/hmcchk-web/pull/979/files ?

tamarayustian:
> code content actually includes both FE and BE changes? duplicate of https://github.com/hmcc-global/hmcchk-web/pull/979/files ?

@ksuhartono97 I just merged the BE and rebased from `release-database`, so now this PR is just 2 files changed

ui/src/components/admin/leadershipTeam/AdminLeadershipTeamContainer.js
```diff
@@ -1,125 +1,368 @@
...
     }
   };
 
-  useEffect(() => {
-    if (api) {
-      if (teams && teams.length) {
-        api.hideOverlay();
-      } else {
-        api.showLoadingOverlay();
+  const duplicateHandler = async (e) => {
```
ksuhartono97, line 144: the duplicate functionality sets the last updated by to be by the original modifier. e.g. if original one says "Tamara" after duplicate it will say "Tamara" too. Should this be changed to the user that initiates the duplicate operation?

ui/src/components/admin/leadershipTeam/AdminLeadershipTeamContainer.js
```diff
@@ -1,125 +1,368 @@
...
+
+        <Button colorScheme="red" onClick={resetHandler} mr={2}>
+          RESET
+        </Button>
+        <Button
+          leftIcon={id !== '' ? <EditIcon /> : <AddIcon />}
+          colorScheme="teal"
+          onClick={onOpen}
+        >
+          {id !== '' ? 'EDIT' : 'ADD'}
```
ksuhartono97, line 240: there's actually an annoying UX here, once you select any row on AgGrid, looks like i can't easily unselect it. Locking the button to forever be in edit state. I would need to refresh to be able to add a new entry if i wish to do so.

ui/src/components/admin/leadershipTeam/AdminLeadershipTeamContainer.js
```diff
@@ -1,125 +1,368 @@
...
+                  <FormLabel>LIFE Group</FormLabel>
+                  <Input
+                    type="text"
+                    value={lifeGroup}
+                    onChange={(e) => setLifeGroup(e.target.value)}
+                  />
+                </FormControl>
+
+                <FormControl>
+                  <FormLabel>Leaders</FormLabel>
```
ksuhartono97, line 323: should this be labeled as leaders' emails and then should we check that these are all emails?

ksuhartono97: since the actual field seems to be called leader's emails in the file below

ksuhartono97 (changes requested):

ui/src/components/admin/leadershipTeam/AdminLeadershipTeamContainer.js
tamarayustian: image
oh yea you have to click the RESET button here to unselect it. i guess that's not very straightforward? or would a description be better or is there a better name for the button?

tamarayustian: note to self: 
- [ ] make the RESET button here more intuitive looking. 
- [ ] also make these buttons to be sticky so you can see it even when you scroll to the bottom of the table

ksuhartono97 (approved):

PR #1011, opened 2024-04-17
1003 leadership team static data

albertparedandan:


ksuhartono97 (changes requested):
small changes only + 1 question: did we want to check the validity period of the leadership team?

albertparedandan:
> small changes only + 1 question: did we want to check the validity period of the leadership team?

already done in the get-latest-leadership

ksuhartono97 (approved):

PR #1018, opened 2024-04-20
GH-996: events UI

cetair:


ui/src/components/events/getWeek.js
```diff
@@ -0,0 +1,15 @@
+export default function isDateInThisWeek(date) {
+  const todayObj = new Date();
+  const eventDate = new Date(new Date(date).toDateString());
+  const todayDate = todayObj.getDate();
+  const todayDay = todayObj.getDay();
+
+  const firstDayOfWeek = new Date(
+    new Date(todayObj.setDate(todayDate - todayDay)).toDateString()
+  );
```
bhryck, line 9: I think doing it like this causes sunday to be the first day of the week so sunday celebration wont show up in this filter. maybe can set Monday as start of week instead

ui/src/components/events/EventCard.js
```diff
@@ -11,25 +12,42 @@ import {
...
-import { useState } from 'react';
+import { FaCalendarAlt } from 'react-icons/fa';
+import { MdOutlineAddToPhotos } from 'react-icons/md';
+import { useState, useEffect } from 'react';
 import { getRenderDate } from '../helpers/eventsHelpers';
 import { DateTime } from 'luxon';
 import { generateGoogleCalendarLink } from '../helpers/eventsHelpers';
 import ReactMarkdown from 'react-markdown';
 import ChakraUIRenderer from 'chakra-ui-markdown-renderer';
+import { ColorExtractor } from 'react-color-extractor';
```
bhryck, line 32: so i tried the cors thing and it didn't work, so i think i can move the color extractor part to the announcement editor. so we just save the color when a file is uploaded instead of rendering it everytime the event page is loaded

https://stackoverflow.com/questions/29670703/how-to-use-cors-anywhere-to-reverse-proxy-and-add-cors-headers
this worked

cetair: can we just put this on hold then

Issue #1012, opened 2024-04-17
Form new sign up alerts


albertparedandan:
allow Forms to specify sign up alert type (Custom, Lifestage, LG, Campus)

PR #1025, opened 2024-04-27
GH-1012: Send email alerts

tamarayustian:


server/api/helpers/forms/get-form-alert-recipients.js
```diff
@@ -25,14 +24,20 @@ module.exports = {
 
     try {
       const formData = await Form.findOne({
-        _id: formId
+        _id: formId,
       });
       // eslint-disable-next-line eqeqeq
       console.log(formData);
```
ksuhartono97, line 30: i think we can remove this console log

server/api/helpers/forms/get-form-alert-recipients.js
```diff
@@ -25,14 +24,20 @@ module.exports = {
...
-      if (formData == null || formData.alertType === 'Custom' || formData.alertType === 'None') return exits.success([]);
+      if (
+        formData === null ||
+        formData.alertType === 'Custom' ||
+        formData.alertType === 'None'
+      )
+        return exits.success([]);
       console.log('loop');
```
ksuhartono97, line 37: and this one too lol

server/api/controllers/forms/post-create-submission.js
```diff
@@ -153,6 +154,71 @@ module.exports = {
...
+              today <= new Date(entry.seasonTo)
+            ) {
+              emailRecipients = entry.leaderEmails;
+            }
+          }
+        }
+
+        // failsafe
+        if (emailRecipients.length <= 0) {
+          emailRecipients = ['web@hongkong.hmcc.net'];
```
ksuhartono97, line 204: do we want to do web@hongkong.hmcc.net? then everyone in t3ch would have gotten it ya haahahah

server/api/controllers/forms/post-create-submission.js
```diff
@@ -153,6 +154,71 @@ module.exports = {
...
+
+          const recipientsEntries =
+            recipientsDict[submissionData[fieldSelected]];
+
+          for (const entry of recipientsEntries) {
+            if (
+              today >= new Date(entry.seasonFrom) &&
+              today <= new Date(entry.seasonTo)
+            ) {
+              emailRecipients = entry.leaderEmails;
```
ksuhartono97, line 197: shouldn't this be an array with append? if there are multiple recipients it will just overwrite again and only send to the last person in list?

server/api/controllers/forms/post-create-submission.js
```diff
@@ -153,6 +154,71 @@ module.exports = {
...
+            formRecord[0].alertType
+          )
+        ) {
+          const recipientsDict =
+            await sails.helpers.forms.getFormAlertRecipients(formId);
+
+          // get the current season leadership teams
+          const today = new Date();
+
+          let fieldSelected;
```
ksuhartono97, line 175: do we need to do error handling here like in case somehow there's bad input or smth.

ksuhartono97 (changes requested):

server/api/controllers/forms/post-create-submission.js
```diff
@@ -153,6 +154,71 @@ module.exports = {
...
+
+        // send notification email
+        await sails.helpers.sendTemplateEmail.with({
+          to: emailRecipients[0],
+          cc: emailRecipients,
+          subject: 'New Sign-up for ' + formRecord[0].formName,
+          template: 'form-notify-leader-success',
+          templateData: {
+            submissionTime: res.createdAt,
+            fullName: user.fullName || submissionData['fullName'],
```
ksuhartono97, line 215: we should do error handling for these fields. if no full name, email, and phone number and user is not logged in, this section would crash

ksuhartono97 (changes requested):

server/api/controllers/forms/post-create-submission.js
```diff
@@ -153,6 +154,71 @@ module.exports = {
         });
       }
 
+      // Send alert email to leaders if setting is turned on
+      if (formRecord[0].alertType !== 'None') {
+        let emailRecipients;
+
+        // get email addresses of the recipients
+        if (formRecord[0].alertType === 'Custom') {
+          emailRecipients = formRecord[0].customAlertRecipients;
```
ksuhartono97, line 163: this part needs to convert the input string to array i suppose

ksuhartono97 (changes requested):

server/api/controllers/forms/post-create-submission.js
tamarayustian: where do we want it to go then ya? because ET getting the failed email doesn't make sense too, because they have to tell us which one failed anyway. or just no need to set a failsafe?

server/api/controllers/forms/post-create-submission.js
tamarayustian: the leaderEmails will already be an array though. the model itself is defined like this:
```
    leaderEmails: {
      type: 'json',
      columnType: 'array',
      defaultsTo: [],
    },
```

ksuhartono97: oh but how come we're looping through the recipientEntries

server/api/controllers/forms/post-create-submission.js
ksuhartono97: should we dump it to no-reply

server/api/controllers/forms/post-create-submission.js
tamarayustian: because the recipientEntries returns a dictionary like this

```
[{
Campus: [{},{}],
Single Adults: [{}, {}],
}]
```

like the function grouped together all the categories and return all the existing entries

server/api/controllers/forms/post-create-submission.js
tamarayustian: i camelized the field selected, does that work in terms of bad input? because the fieldSelected will always be one of the choices of the Alert Types and then camelize that

ksuhartono97 (approved):

PR #1045, opened 2024-05-26
GH-1033: changed design

Baeck016:


bhryck (approved):
if crash not my fault

PR #1050, opened 2024-06-06
1049 sermon editor bugs fixes

Midnite1212:
Bugs List
- [x] Sort the sermon list by most recent dates
- [x] add the sn to live sermon
- [x] fix esv.json not uploading to uat
- [x] fix text styling and buttons in editor
- [x] fix broken hamburger menu
- [x] fix undefined content crash for user sn
- [x] add public link to the sn list
- [x] add a save button for the sn to save in cookies instead
- [x] user sn only saves first one
- [x] add a losding screen when sermon not fetched yet
- [ ] ~have to enter the bible verse to detect the bible verse node~
- [x] specific bug in bible verse extension, cannot detect verses with diff number of digits (e.g. READ James 5:9-10)
- [x] sn content doesnt show up in mobile


alvone01 (approved):

ui/src/components/sermon-notes/SermonNotesContainer.js
```diff
@@ -50,7 +53,7 @@ const SermonNotesContainer = (props) => {
         const { data, status } = await axios.put(
           '/api/user-sermon-notes/update',
           {
-            userId: user.id,
+            userId: user?.id || '',
```
Baeck016, line 56: Is there a scenario when the user-sermon-notes doesn't contain userId?

ui/src/components/sermon-notes/SermonNotesContainer.js
```diff
@@ -131,13 +138,13 @@ const SermonNotesContainer = (props) => {
       return sermonNotes?.originalContent;
     }
   }, [userSermonNotes, sermonNotes?.originalContent]);
-
+  console.log(sermonNotes);
```
Baeck016, line 141: Delete this part

ui/src/components/sermon-notes/SermonNotesContainer.js
Midnite1212: yes, that function is called on mount. But if the user is guest or no account, they would not have user object, n if we pass null to the API it'll crash lol. better give a dummy string

PR #1052, opened 2024-06-08
GH-1051: Made fixes for Toast and Duplicate Logic

JoshNg123:


ui/src/components/admin/sermonNotes/SermonNotesEditorModal.js
```diff
@@ -107,6 +107,32 @@ const SermonNotesEditorModal = (props) => {
     }
   }, [toast]);
 
+  const checkDuplicateTitleAndSermonLink = async () => {
+    try {
+      const data = await fetchSermonNotes();
```
Midnite1212, line 112: you shouldn't be fetching the data again here, since you already fetch it in the initial mount. Ideally we should minimize the API calls to reduce network.

Midnite1212 (changes requested):

Midnite1212 (approved):
lgtm 👍

PR #1057, opened 2024-06-23
1056 minor fixes

albertparedandan:


server/api/helpers/sermons/get-sermons.js
```diff
@@ -1,6 +1,22 @@
...
+const parseSermonTitle = (title) => {
+  // eslint-disable-next-line eqeqeq
+  if (title == null) return '';
+
+  const decodedTitle = he.decode(title);
+  const pattern = /.*(Part \d+.*)/g;
+  const captured = pattern.exec(decodedTitle);
+
+  // eslint-disable-next-line eqeqeq
+  return captured != null && captured[1] != null && captured[1] !== '' ? captured[1] : decodedTitle;
```
bhryck, line 17: if we edit in the api like this, then the title in this page, the title will just be "Part 3: Joy", I think we should display the full title in this case
image

bhryck: so i can think of 2 ways to do this:
- add an argument in the API to return the parsed title or the full title
- add the 'series' back into the title in this page, but i think this is hacky

albertparedandan: i think if others agree with this suggestion, the solution is just to add a new field `fullTitle` or something. so card will use `sermon.title` while sermon pages will use `sermon.fullTitle`

bhryck (approved):

PR #1066, opened 2024-06-28
GH-1065: fixing fill in blanks for sn

Baeck016:


ui/src/components/helpers/TipTap/FillInBlank.js
```diff
@@ -27,16 +34,12 @@ export const FillInBlankNode = Node.create({
           const position = editor.state.selection.from;
           return commands.insertContentAt(position, {
             type: 'fillInBlank',
-            attrs: { editorText: fillText, userText: '' },
+            attrs: { editorText: fillText, userText: '', currentId: uuid() },
```
Midnite1212, line 37: fyi, the currentId here will be the same for all the fillIntheBlanks node, because the way tiptap works, it will render this command only once, and keep calling the same render when u input fill in the blank.

U need to force it to rerender so the id change for each fill in the blank, can look into the `addProseMirrorPlugin()` method I use in notes.

Midnite1212 (changes requested):

ui/src/components/helpers/TipTap/FillInBlank.js
Baeck016: I know what you mean, bc Node will only create once with the same attribute across the instances. But by doing what I have done with, it will create different uuid for each instances

Midnite1212 (approved):

bhryck (approved):

Midnite1212 (approved):

PR #1071, opened 2024-07-04
GH-1: fix sn save note

Baeck016:


ui/src/components/sermon-notes/SermonNotesContainer.js
```diff
@@ -257,6 +267,7 @@ const SermonNotesContainer = (props) => {
               />
             )}
             <Button
+              display={isLogin ? 'button' : 'none'}
```
Midnite1212, line 270: I dont think display ='button' actually exist? can do block / flex instead.

ui/src/components/sermon-notes/SermonNotesContainer.js
```diff
@@ -163,6 +164,15 @@ const SermonNotesContainer = (props) => {
...
 
+  useEffect(() => {
+    if (!user?.id) {
+      setLogin(false);
+    } else {
+      setLogin(true);
+    }
+    console.log(user?.id);
+  }, [user]);
+
```
Midnite1212, line 175: [minor]: the website will auto reload if there is a user, so u can just check directly whether user.id exist or not and show the button or not. This also works but i think its a bit overkill w useEffect lol.

ui/src/components/sermon-notes/SermonNotesContainer.js
```diff
@@ -21,6 +21,7 @@ const SermonNotesContainer = (props) => {
   const [userSermonNotes, setUserSermonNotes] = useState();
   const [editUserSermonNotes, setEditUserSermonNotes] = useState();
   const toast = useToast();
+  const [isLogin, setLogin] = useState(false);
```
Midnite1212, line 24: Since u change the way the button is rendered, i think this is unused? Can check n delete along w the useEffect if its unused.

Midnite1212 (approved):
🎉🎉

alvone01 (approved):

PR #1089, opened 2024-07-22
1079 web vision mission

bhryck:


ui/src/components/home/VisionMissionSection.js
```diff
@@ -0,0 +1,412 @@
...
+                    >
+                      VISION
+                    </Text>
+                  </Box>
+                  <Box>
+                    <Text>
+                      <Text display="inline">
+                        <Text
+                          color="#4A6EEB"
+                          fontSize="20px"
```
ksuhartono97, line 90: i'm pretty sure you can move the "fontSize" "fontWeight" "lineHeight" to the parent element and it would still apply the effect for the child

ui/src/components/home/VisionMissionSection.js
```diff
@@ -0,0 +1,412 @@
...
+                    lineHeight="1"
+                    wordBreak="break-word"
+                  >
+                    Our Vision<br></br> & Mission
+                  </Heading>
+                </Box>
+              </Box>
+              <Flex>
+                <VStack
+                  borderRadius="0.5rem"
```
ksuhartono97, line 65: for paddings and border radius, using the built in chakra reactive numbers is usually more reactive towards ui change.

i.e. `borderRadius="10"`

Also the padding and border looks different from design.

Instead of `1em` I think you can do `p=10`

ksuhartono97: image
design padding is a lot bigger

image


similar problem for most paddings

ui/src/components/home/VisionMissionSection.js
```diff
@@ -0,0 +1,412 @@
...
+                    <Text
+                      color="#0C0C20"
+                      fontSize="20px"
+                      fontFamily="Manrope"
+                      fontWeight="700"
+                      textTransform="uppercase"
+                      lineHeight="20.6px"
+                      letterSpacing="4px"
+                      wordBreak="break-word"
+                    >
```
ksuhartono97, line 81: need padding
image

ui/src/components/home/VisionMissionSection.js
ksuhartono97: also 'fontSize' is usually better to use rem and then try to eyeball match with the design. makes it better for different device sizes i think. a font size of 20 in Figma was a match with `1.4rem` for me. So I think this current fontSize is too small. You can somewhat do mathematical conversion using this reference point to match font sizes on the design with a bit of tuning (rounding up/down)

ui/src/components/home/VisionMissionSection.js
```diff
@@ -0,0 +1,412 @@
...
+                          churches in
+                        </Text>{' '}
+                        <Text
+                          color="#141419"
+                          fontSize="20px"
+                          fontFamily="DMSerifDisplay_Italic, serif"
+                          fontWeight="400"
+                          lineHeight="28.84px"
+                          wordBreak="break-word"
+                          display="inline"
```
ksuhartono97, line 117: You can actually do 

```
<Text
as="span"
>
....
</Text>

```

This will allow the parent element to directly handle it as an individual text. So you don't need to put display "inline" If you are finding that it doesn't render spaces properly, you can put

`whiteSpace="pre-wrap"` to make it follow your whitespaces,

ui/src/components/home/VisionMissionSection.js
```diff
@@ -0,0 +1,412 @@
...
+                      </Button>
+                    </Link>
+                  </Flex>
+                  <VStack bgColor="white">
+                    <Box textAlign="left" w="100%" p="1em">
+                      <Text
+                        color="#4A6EEB"
+                        fontSize="20px"
+                        fontFamily="Manrope"
+                        fontWeight="400"
```
ksuhartono97, line 378: fontWeight should be 700

ui/src/components/home/VisionMissionSection.js
```diff
@@ -0,0 +1,412 @@
...
+                        my="auto"
+                        borderRadius="2.75rem"
+                        _hover={{ bg: '#4A6EEB', color: 'white' }}
+                        textColor="#4A6EEB"
+                        bgColor="transparent"
+                      >
+                        <Text
+                          fontSize="18px"
+                          fontFamily="Manrope"
+                          fontWeight="400"
```
ksuhartono97, line 363: fontWeight should be 700

ui/src/components/home/VisionMissionSection.js
ksuhartono97: also padding and borderRadius is wrong

ui/src/components/home/VisionMissionSectionMobile.js
```diff
@@ -0,0 +1,373 @@
...
+                      locally, regionally, and globally for the spread of
+                      Jesus's fame.
+                    </Text>
+                  </VStack>
+                </VStack>
+                <Link href="/about-us" id="homepage-about">
+                  <Button
+                    my="auto"
+                    borderRadius="2.75rem"
+                    w="100%"
```
ksuhartono97, line 346: button not actually stretchign to fit the container on mobile

image

ksuhartono97 (changes requested):
Paddings, border radius, and text weight appear different from design on web version. Mobile version looks ok!

ksuhartono97 (approved):

PR #1094, opened 2024-08-12
1 sn hotfix

Midnite1212:
- [x] Add sn link shortener functions to domain
- [x] Fix sn can be put at any level now with correct functionality and width.


ksuhartono97 (approved):
lgtm

PR #1096, opened 2024-08-16
GH-1091: #3

albertparedandan:
Added function to send query emails for single/batch user data update.


server/api/helpers/parseUserQuery/parse-user-query.js
```diff
@@ -0,0 +1,242 @@
...
+      }
+    }
+  } else {
+    const userData = data[3];
+    if (userData.length === 4 && verifyUserData(userData)) {
+      users.push({
+        name: userData[0],
+        phoneNumber: userData[1],
+        email: userData[2],
+        lifestage: userData[3]
```
ksuhartono97, line 49: i think we said we need Campus as well? I think this will be important for DB as well as we would need that as demographic data.

albertparedandan: We can, but I can imagine it being very annoying for UST  leaders (or any campus leaders) to have to fill every single line with the same thing

ksuhartono97: yeah idk what would be a better workflow, do we just force set it from the backend? like this LG is defined as under a certain campus... so all it's members are set to be part of the campus. but not sure if this is a good/complete solution. e.g. if there are people that are part of "Others" LG

albertparedandan: Yeah, wouldn't work in summer either. I think potential solution is just to require them to fill the campus column and also for u to create the LG directory template which also requires members to fill in their campus.

That way leaders just have to copas

ksuhartono97 (approved):

PR #1099, opened 2024-08-21
1080 experience community

bhryck:


ksuhartono97:
image
think this PR should also just remove these FABs

ui/src/components/home/ExperienceBiblicalCommunity.js
```diff
@@ -0,0 +1,468 @@
...
+
+const ExperienceBiblicalCommunity = () => {
+  const [isMobile] = useMediaQuery('(max-width: 768px)');
+  const [isOpen, setIsOpen] = useState(false);
+  const onClose = () => setIsOpen(false);
+  const onOpen = () => setIsOpen(true);
+  const settings = {
+    infinite: true,
+    autoplay: true,
+    autoplaySpeed: 3000,
```
ksuhartono97, line 34: in the original design demo this carousel was continuously moving? this demo is quite slow the movement. not sure if this is what the designers wanted, maybe can check with them. if they want it to continuously move, i think you just need to make the `speed` larger than the `autoplaySpeed`

also `css-ease:linear` will make the speedup and slowdown more smooth

ui/src/components/home/ExperienceBiblicalCommunity.js
```diff
@@ -0,0 +1,468 @@
...
+                      >
+                        <Button
+                          w="100%"
+                          pl={10.5}
+                          pr={7}
+                          borderRadius={80}
+                          border="0.70px solid #D46764"
+                          justify="space-between"
+                          align="center"
+                          bgColor="#D46764"
```
ksuhartono97, line 205: image

gray background in button. this `bgColor` should be nothing/same as container background. also bigger arrow (use `boxSize`) and also wider gap between the button contents (can increase button width i think?)

image

ksuhartono97: image
also the button has a weird gray effect after being clicked on mobile.

i suspect you need to add a custom `_hover={{css:}}`. can see the chakra docs for this/see my events PR

ksuhartono97 (changes requested):

ui/src/components/home/ExperienceBiblicalCommunity.js
bhryck: changed the speed, think maybe best to release to uat first to see which one design team prefers

ksuhartono97 (approved):

PR #1112, opened 2024-09-01
GH-1: automate updating live page

Midnite1212:


albertparedandan:
Gonna decline this. I think it makes more sense to do the opposite given snotes are usually late and we want the online sermon up first. So it should be updating live sermon creates a new template for new sermon notes instead of the opposite (what this PR is doing).

PR #1119, opened 2024-09-25
GH-1118: Switch from u/p to oauth2 for gmail funcs

albertparedandan:


albertparedandan:
have to update .env before merging

ksuhartono97 (approved):

PR #1127, opened 2024-10-19
GH-1026: Offline Sermon Notes in Profile and Sermon Page

albertparedandan:


server/api/controllers/sermons/get-sermon-notes.js
```diff
@@ -0,0 +1,66 @@
...
+      }).sort('date DESC');
+
+      // Step 2: For each parent sermon note, if userId is provided, check for corresponding user sermon notes
+      let data = await Promise.all(
+        parentSermonNotes.map(async (parentNote) => {
+          let isSaved = false;
+          let childUpdatedAt = null;
+
+          if (userId) {
+            let userSermonNote = await UserSermonNotes.findOne({
```
albertparedandan, line 39: this is very inefficient. you're calling the DB on every parentSermonNotes. i would simplify the query and only call the DB once. 

```
const allRelatedUserSermonNotes = await UserSermonNotes.find({
  userId : userId,
  isDeleted: false,
  sermonId: parentSermonNotes.map(i => i.sermonId)
});
```

afterwards, you can enrich each parentNote with the userSermonNote

PR #1134, opened 2024-11-15
GH-1133: Add Retreat emails

albertparedandan:


server/views/emails/form-retreat-success.ejs
```diff
@@ -57,14 +64,12 @@
       <li>LATITUDE</li>
     </ul>
   </li>
-  <li>Example Remark Note: RT: Bob Chan INDIVIDUAL</li>
+  <li>Example Remark Note: RT: Bob Chan [INDIVIDUAL]</li>
```
ksuhartono97, line 67: the remark on the email example is [Early Registration]

PR #1136, opened 2024-11-26
GH-1135: Migrate Fundraise

JoshNg123:


server/api/controllers/fundraise/get-fundraise.js
```diff
@@ -0,0 +1,31 @@
...
+    try {
+      if (campaignName) {
+        const data = await Fundraise.find({
+          campaignName,
+          isDeleted: false,
+        });
+
+        return exits.success(data);
+      }
+      return exits.invalid('Missing required field: campaignName');
```
albertparedandan, line 25: Minor, can leave it if no time - but this is not necessary as you've made the parameter required above

albertparedandan (approved):

server/api/controllers/fundraise/get-fundraise.js
JoshNg123: Done

PR #1167, opened 2025-03-07
1165 discover home

Baeck016:


cornyachtsorry:
- iPad looks off
image

- Hover doesn't work
image

PR #1181, opened 2025-04-02
1179 easter hero

bhryck:


ui/src/components/events/EventsExport.js
ksuhartono97, line 1: How come this file is in this PR ya?  Did we.. somehow remove this change? I also don't see it being used haha.

ui/src/components/events/EventsPage.js
ksuhartono97, line 1: Would we want to separate this into a PR to main instead so the changes are kept? Unless we integrate the changes from the Easter Branch forever to our main branch, which I feel like is not the best since we change the navbars and everything.

bhryck: leme raise another PR to main. can keep this change in this PR too

ui/src/components/events/EventsExport.js
bhryck: my bad, prob forgot to reset the branch. removed

PR #1204, opened 2025-05-08
GH-2303: sn bug duplicate id

bhryck:


albertparedandan:
I don't think this is the right way to do it, but will approve for now given it's legacy logic n we're just trying to fix the bug

albertparedandan (approved):

Issue #1206, opened 2025-05-21
BB Site - Curriculum Section


cornyachtsorry:
![Image](https://github.com/user-attachments/assets/705eab9f-33b2-4044-9298-8d754c93f32f)

cornyachtsorry:
LGTM

PR #1210, opened 2025-05-29
GH-1206: BB Curriculum

cornyachtsorry:


ui/src/components/building-block/Curriculum.js
```diff
@@ -0,0 +1,209 @@
...
+import {
+  Box,
+  Flex,
+  Image,
+  Text,
+  Heading,
+  IconButton,
+  HStack,
+  VStack,
+  useBreakpointValue,
```
Baeck016, line 11: delete library not being used.

Baeck016 (approved):
delete lib not being used

ui/src/components/building-block/Curriculum.js
cornyachtsorry: resolvedd

Issue #1219, opened 2025-06-29
About Page - Vission, Mission & Values and Beliefs


kuritsuan:
<img width="724" alt="Image" src="https://github.com/user-attachments/assets/da06d1be-5d53-448f-90f9-b8663e98a5a3" />

<img width="724" alt="Image" src="https://github.com/user-attachments/assets/89932ffd-b29e-415d-a8a1-b963489011c2" />

PR #1221, opened 2025-07-03
GH-1219 about vision beliefs

kuritsuan:


Baeck016:
Please delete the files that you have replaced and won't be using anymore, cuz i see alot of belief smth files.

PR #1223, opened 2025-07-03
1217 about ourstrategy

cornyachtsorry:


Baeck016:
image
there is a bg color

Baeck016:
image
I feel like all the sub-header can be bolded too

PR #1271, opened 2025-10-02
1267 seo structured data

bhryck:


ui/src/components/seo/seoData.js
```diff
@@ -0,0 +1,155 @@
+// Consolidated SEO configuration for meta tags
+export const seoData = {
```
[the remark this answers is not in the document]
albertparedandan: we should probably have ET/ at least Kris review all the descriptions here given it's public facing

ui/src/components/seo/seoData.js
```diff
@@ -0,0 +1,155 @@
...
+  if (pathname.startsWith('/sermons/')) {
+    return {
+      title: 'Sermon | HMCC Hong Kong',
+      description:
+        'Listen to this sermon from Harvest Mission Community Church Hong Kong. Biblical teaching and spiritual encouragement for your faith journey.',
+      keywords: 'sermon, preaching, bible teaching, HMCC Hong Kong',
+    };
+  }
+
+  if (pathname.startsWith('/witness')) {
```
albertparedandan, line 121: is this still relevant/do we want to include these?

ui/src/components/seo/seoData.js
```diff
@@ -0,0 +1,155 @@
...
+    return {
+      title: 'Testimonies | Witness Stories | HMCC Hong Kong',
+      description:
+        'Personal testimonies and faith stories from Harvest Mission Community Church Hong Kong community members. Be encouraged by how God is working.',
+      keywords: 'testimonies, witness stories, faith stories, HMCC Hong Kong',
+    };
+  }
+
+  if (pathname.startsWith('/admin')) {
```
albertparedandan, line 130: we should take this out. it's not something we want SEO to crawl i think

ui/src/components/seo/seoData.js
```diff
@@ -0,0 +1,155 @@
...
+    return {
+      title: 'Admin | HMCC Hong Kong',
+      description:
+        'Church administration portal for Harvest Mission Community Church Hong Kong staff and leadership.',
+      keywords: 'admin, church administration, HMCC Hong Kong',
+    };
+  }
+
+  if (pathname.startsWith('/forms/')) {
```
albertparedandan, line 139: this isn't a valid link, so i don't think we need to include these?

ui/src/components/seo/SchemaGenerator.js
```diff
@@ -0,0 +1,428 @@
+export const generateChurchSchema = (pathname = '/') => {
```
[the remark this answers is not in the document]
albertparedandan: same comment as the seoData.js

bhryck:
@albertparedandan , i removed the description for the pages we dont want public. but i feel like the right way to do it is to add no-index to those pages, i wonder if GSC has some feature to do that

ui/src/components/sermons/NoOnlineStreamPage.js
```diff
@@ -0,0 +1,84 @@
...
+              <Heading
+                as="h1"
+                size="xl"
+                fontFamily="DMSerifDisplay_Italic"
+                color="blue.600"
+              >
+                No Live Stream Available
+              </Heading>
+
+              <Text fontSize="lg" color="gray.600" maxW="md" lineHeight="tall">
```
albertparedandan: change the order of text below? i think this makes more sense:

No Live Stream Available
Our live streams typically happen....

You can access our past sermons...
Button to browse past sermons

PR #1308, opened 2026-01-31
1299 shine giving

Baeck016:


Baeck016 (commented):
There is no mobile version?

PR #1330, opened 2026-03-08
GH-1327: Easter Banner

Triangulr:


ui/src/components/easter/EasterContainer.js
```diff
@@ -11,7 +11,79 @@ const EasterContainer = () => {
...
       fontFamily={theme.fonts?.body}
     >
-      <Container maxW="container.lg">
+      <Container maxW="container.xl" px={{ base: 6, md: 10 }}>
+        <Heading
+          as="h1"
+          fontFamily="'DM Serif Display', Georgia, serif"
+          fontStyle="italic"
+          fontWeight="400"
+          fontSize={{ base: '4xl', lg: '60px' }}
```
Midnite1212: it's better to streamline the units for fontSizes, spacing etc. Same for the other parts where you use different units.

Also, we are slowly transitioning to use em instead of px, so its more responsive instead of px's absolute value.

bhryck (approved):

PR #1342, opened 2026-03-15
GH-1339: check it out button

bhryck:


ui/src/components/home/HeroSection.js
```diff
@@ -396,6 +431,125 @@ const HeroSection = () => {
...
+                borderRadius="80px"
+                display="inline-flex"
+                justifyContent="center"
+                alignItems="center"
+                 transition="transform .18s ease, background-color .18s ease, box-shadow .18s ease"
+                _hover={{
+                  bg: "linear-gradient(90deg, #6B1F38 0%, #D67E8E 100%) ",transform: 'translateY(-3px)', boxShadow: 'lg'
+                }}
+                _active={{
+                  bg: "linear-gradient(90deg, #5A1A2F 0%, #C26A7D 100%)"
```
Midnite1212: I think this whole section no need to separate between mobile and desktop. Instead of copy pasting the flex with different breakpoints, better to use this breakpoints in the fontsizes, padding etc. that is different between mobile and desktop for cleaner and maintainable code.

PR #1393, opened 2026-06-28
[Class Tracker] Full Implementation

yuenyauu:


specs/admin-class-forms-tracking.md
```diff
@@ -0,0 +1,99 @@
...
+  muted text, their header carries a tooltip explaining where to change them, and double-clicking
+  one shows an info message pointing the admin to the Form Editor.
+- **Archived courses** render in muted text and are **collapsed by default** — only their Status
+  column shows, with an expand toggle to reveal the rest. Their data is read-only.
+
+## Data model
+
+Two layers — a mutable template on the form, an immutable snapshot per registrant.
+
+**Form template** (`Form.isClass`, `Form.courses[]`): each course is
```
paretypus: i think instead of Form.courses[] we should do Form.classTrackingTempalate which will be a json containing courses[] as we will have other column types in the future, so i wanna keep them in the same object.

feel free to ask Claude what's a better field name for this

Plans/admin-class-forms-tracking.md
```diff
@@ -0,0 +1,99 @@
...
+registrant's progress through every course (status, start/completion dates, remarks) is
+tracked and edited inline from the existing admin "View Data" submission grid.
+
+The one tricky constraint: course lineups change between seasons, and **a registrant's
+historical record must never be rewritten by a later config change**. We handle this by
+snapshotting each registrant's courses at submission time (see Data model below).
+
+## Workflows
+
+**Setting up a class** — In the form editor, toggle "Is this a class?" and add courses
```
paretypus, line 17: instead of adding another toggle (as it's getting quite crowded in the form creator page), can we do a dropdown with header "Tracking Type" with values: Payment, Class, None

if "Payment" selected, then show the fields that's currently being shown if payment required is toggled on
if "Class" selected, then show the fields related to class tracking column

server/api/controllers/forms/post-create-submission.js
```diff
@@ -40,7 +40,7 @@ module.exports = {
         isPublished: true,
       });
 
-      if (formRecord === null) {
+      if (formRecord.length === 0) {
```
paretypus, line 43: how come this becomes null unsafe

server/api/models/Form.js
```diff
@@ -78,5 +78,16 @@ module.exports = {
     externalFormLink: {
       type: 'string',
     },
+    isClass: {
+      type: 'boolean',
+      defaultsTo: false,
+    },
+    courses: {
```
paretypus: see my comment about classTrackingTempalate

server/api/controllers/classTrackingData/update-class-tracking-data.js
```diff
@@ -0,0 +1,114 @@
...
+      if (!record) {
+        return exits.invalid('Class tracking record not found');
+      }
+
+      const courseExists = record.courses.some(
+        (course) => course.courseId === courseId
+      );
+      if (!courseExists) {
+        return exits.invalid('Course not found on this record');
+      }
```
paretypus, line 113: can we do another check here to check if course is inactive or not? just to block user from updating an inactive course tracking column

ui/src/components/forms/FormEditor.js
```diff
@@ -176,6 +176,8 @@ const FormEditor = (props) => {
           formInformation.paymentConfirmationEmailTemplate,
         paymentEmailSubject: formInformation.paymentEmailSubject,
         paymentCcEmail: paymentCcEmail,
+        isClass: formInformation.isClass,
+        courses: formInformation.courses,
```
paretypus: same comment about classTrackingTemplate

ui/src/components/forms/FormEditorContainer.js
```diff
@@ -174,6 +202,28 @@ const FormEditorContainer = (props) => {
...
   };
 
+  const addCourse = () => {
+    setCourses([...courses, newCourse()]);
+  };
+
+  const removeCourse = (courseId) => {
+    if (
+      window.confirm(
+        'Are you sure? If this course already has registrant tracking data, the server will reject this removal — deactivate the course instead.'
```
paretypus: i don't think we should even have this option if server is going to reject and auto convert to deactivate. we should just allow users to deactivate instead of remove course

paretypus (changes requested):

server/api/controllers/forms/post-create-submission.js
yuenyauu: `Form.find()` (Waterline) always resolves to an array, even when nothing matches — it never returns `null`. So the old `formRecord === null` check could never be true, and an invalid `formId` would silently fall through to `formRecord[0].isPaymentRequired` / `formRecord[0].isClass` later, where `formRecord[0]` is `undefined` — that's the actual null-unsafe path. Switching to `formRecord.length === 0` fixes that: it correctly guards against an empty result and returns `exits.invalid()` before we ever touch `formRecord[0]`.

ui/src/components/forms/FormEditorContainer.js
yuenyauu: Done. "Delete" is only shown on initial class form creation -- i.e., when courses have no tracking data yet.

Plans/admin-class-forms-tracking.md
yuenyauu: I used Tabs instead, not dropdown.

specs/admin-class-forms-tracking.md
yuenyauu: Done.

server/api/controllers/classTrackingData/update-class-tracking-data.js
yuenyauu: The active/inactive state is tracked on the Form class, not the ClassTrackingData class -- youc an see the check happening in the later lines.

```
const courseExists = record.courses.some(
  (course) => course.courseId === courseId
);
if (!courseExists) {
  return exits.invalid('Course not found on this record');
}

const form = await Form.findOne({ id: record.formId });
const templateCourse = (
  (form && form.classTrackingTemplate
    ? form.classTrackingTemplate.courses
    : []) || []
).find((course) => course.courseId === courseId);
if (templateCourse && !templateCourse.isActive) {
  return exits.courseInactive(
    'Cannot update tracking data for an archived course'
  );
}
```

server/api/controllers/paymentData/post-send-payment-email.js
```diff
@@ -93,11 +93,10 @@ module.exports = {
 
       if (res) {
         const modelName = `paymentData-${res[0].formId}`;
-        existing = await LastUpdated.updateOne({ modelName })
-          .set({
-            lastUpdatedBy: this.req.user.fullName,
-          })
-          .fetch();
+        // No .fetch() - updateOne() already returns the affected record.
```
paretypus, line 96: can we regression test this just in case

ui/src/pages/admin/forms/classTrackingColumns.js
```diff
@@ -0,0 +1,150 @@
+const classStatusList = ['Not Started', 'In Progress', 'Completed'];
```
paretypus, line 1: i feel like having these constants at 2 different places is not great (here and in the update-class-tracking-data.js). i think we should consider putting these as constants in the config file and creating an endpoint that returns these values.

refer to `formAlertTypes`

ui/src/pages/admin/ag-grid-editors/CustomDateEditor.js
```diff
@@ -45,18 +49,11 @@ export default forwardRef((props, ref) => {
         }
         return dateString;
       },
+      // Nothing was picked, so there is no edit to commit - cancel instead of
+      // writing null over the existing value.
       isCancelAfterEnd: () => {
         return !selectedDate;
       },
```
paretypus, line 56: why is the below removed ya? can we double check this functionality to make sure no regression

ui/src/pages/admin/forms/AdminFormDataViewer.js
```diff
@@ -220,28 +239,42 @@ export default function AdminFormDataViewer(props) {
   const checkIfUpdated = useCallback(
     async (updateData = true) => {
       try {
-        const modelName = `paymentData-${formId}`;
-        const { data } = await axios.get('/api/last-updated', {
-          params: { modelName },
-        });
-        const dateObj = DateTime.fromISO(data);
-        if (!lastUpdatedTime.current || dateObj > lastUpdatedTime.current) {
+        const modelNames = [];
```
paretypus, line 242: similar comment about regression testing paymentData form

paretypus (changes requested):
approve on the condition of regression test.

paretypus (approved):

PR #1402, opened 2026-07-06
GH-1401: Consolidate UI into a single components library (Chakra boundary)

Midnite1212:
Closes #1401. Supersedes #1400 (same work rebased onto current `main` — see "Relation to #1400" below).

## What this PR does

Consolidates all reusable UI into a single components library at `ui/src/components/`, with a hard boundary around Chakra UI. App code no longer imports `@chakra-ui/react` anywhere — everything comes from `components`. This is **structure only**: no visual or behavioral changes, no dependency changes, Chakra stays on v1.

**Why:** every UI-library change used to mean touching ~160 files (see the stalled `1277-*` / `1356-migrate-chakra-v3` branches). With this boundary in place, the planned **Tailwind migration** happens inside one folder while the rest of the app stays untouched.

### The library

```
ui/src/
  components/
    chakra/index.js   ← the ONLY file that imports @chakra-ui/react (star re-export + override hook)
    icons.js          ← the ONLY file that imports @chakra-ui/icons
    index.js          ← the barrel: explicit named re-exports (93 Chakra primitives/hooks + app components)
    README.md         ← the rules
    Footer/  NavigationBar/  TipTap/
    Card.js  FileUpload.js  LiveButton.js  SermonSeries.js  BibleVerseAccordion.js
    Pagination.js  AudioEmbed.js  VideoEmbed.js  parseDescription.js
  utils/              ← non-UI shared logic split out of pages/helpers
                        (customAxios, events/forms/user helpers, lists, Scroll*, SermonNotes, DailyBrp,
                         arrayToExcel, TrackingUtil — consolidated from src/util/)
  pages/              ← routed pages only, incl. seo/ (MainContainer + PrivateRoute stay here: they import pages)
```

### Changes

- **Moved former `components/` content into `pages/`** (they were actually pages, including the new `seo/` folder) and committed it as the base checkpoint; fixed the 3 broken imports in `App.js`.
- **New `ui/jsconfig.json`** (`baseUrl: "src"`) → absolute imports: `import { Box, Button, Footer } from 'components'`. **Restart your dev server after pulling.**
- **~160 files swept**: `'@chakra-ui/react'` → `'components'`, `'@chakra-ui/icons'` → `'components/icons'` (specifier-only rewrites; also caught 4 stray `@chakra-ui/layout`/`select` imports).
- **Relocated cross-page shared UI** into the library and **non-UI helpers** into `src/utils/` — done as pure `git mv` commits followed by separate import-rewrite commits, so file history is preserved (moves show as renames).
- **ESLint guard** (`no-restricted-imports`): any direct `@chakra-ui/*` import outside the library is a lint/build error, so the boundary can't erode.
- **Dead code removed** (verified zero importers): `pages/visit-us/` (the live route uses `discover/PlanAVisit.js`), `pages/connect-prayer/`, empty `pages/Card/`, `src/Plans/`, `src/util/`.
- **Docs**: `ui/src/components/README.md` (library rules) + `AGENTS.md` updated.

### Relation to #1400

Identical work, transplanted commit-by-commit onto current `main` for the new issue-tracked branch flow. Differences that arose from the newer base, all resolved:
- `10y-commitment` pages are **not** in this PR (they belong to `release-10y-commitment`; when that merges, its components land under `pages/` — a trivial follow-up move if it merges first).
- `about/StorySection.js` stays deleted, `StrategySection` → `TransformationPathwaysSection` rename kept.
- Files added on `main` since the original PR (`seo/`, `about/StaffMember.js`, `sermons/NoOnlineStreamPage.js`) were brought into the library conventions in a follow-up commit.

### Design notes for reviewers

- The barrel uses **explicit named exports, never `export *` from two sources** — the ES spec silently drops colliding names from double star-exports (undefined at runtime, no build error). With explicit exports, a future name collision (e.g. our `Card`) fails the build loudly.
- Files **inside** the library import from `./chakra` directly, never from the `components` barrel — avoids module-initialization cycles.
- Consumers of app-owned components may import either the barrel (`import { Footer } from 'components'`) or the module path (`import FileUpload from 'components/FileUpload'`); both are inside the library folder.

### Verification

- `rg "@chakra-ui" ui/src` → matches only in `components/chakra/index.js` + `components/icons.js`
- Production build passes; static barrel check confirms every re-exported name exists in Chakra
- Live render check on `/`, `/sermons`, `/about-us` — no new console errors vs baseline
- Diff audit: moved files changed **only** in import lines; `package.json` diff is the ESLint rule only; no lockfile changes
- Library folder is byte-identical to the reviewed #1400 branch (`git diff refactor/components-library..HEAD -- ui/src/components ui/src/utils` is empty)
- Known pre-existing (NOT from this PR): `npm test` fails on `CurrentSermon.js`'s `import { DATE_FULL } from 'luxon/src/impl/formats'` (untranspiled ESM) — fails identically on `main`; worth a separate one-line fix

## Tailwind migration plan (follow-up PRs, not this one)

The library is the migration ramp. Because every Chakra primitive reaches the app through `components`, the migration is incremental and consumer-invisible:

1. **Set up Tailwind** alongside Chakra (separate PR): `tailwind.config.js`, PostCSS, map the `theme.js` design tokens (DM Sans/Inter, colors, textStyles) into the Tailwind theme.
2. **Replace primitives one at a time inside the library**: add e.g. `components/primitives/Box.js` (Tailwind-based, same props surface we actually use) and switch the barrel's `export { Box } from './chakra'` to the new implementation. The 160+ consumer files don't change. Usage frequency gives the order: `Box` (124 files) → `Text` (107) → `Button` (73) → `Heading` (71) → `VStack`/`Image`/`Flex`/`Container`/`Stack` → long tail.
3. **Migrate app-owned components** (Footer, NavigationBar, Card, …) to Tailwind classes directly — they're all in one folder now.
4. **Interactive/overlay components last** (Modal, Drawer, Accordion, Tabs, useToast/useDisclosure) — these carry behavior + a11y, likely replaced with Headless UI/Radix + Tailwind rather than hand-rolled.
5. **Delete** `@chakra-ui/*`, `@emotion/*`, `framer-motion` (Chakra's dep) from `package.json` when `components/chakra/index.js` has no remaining star re-export — the ESLint guard guarantees nothing outside the library still needs them.

Each step ships as a small, independently-revertable PR; the app renders identically throughout because the barrel's export names and props stay stable.

---

## Regression test plan

This PR claims **"structure only — no visual or behavioral changes."** Because the change is *moves + re-exports*, almost the entire risk surface is verifiable **statically**, before you ever open a browser. The plan is four automated gates plus one runtime smoke pass. Do them in order; stop at the first failure.

**Before you start:** the migration adds `jsconfig.json` (`baseUrl: "src"`), so **restart your dev server / IDE** after checking out. Run everything against a **clean working tree** (`git status` clean — untracked files under `src/components/` will trip the boundary check and the build).

### Gate 0 — Automated harness (30 s, no dev server)

A dependency-free script ships in this PR at `ui/scripts/verify-migration.js`. It tries to *falsify* the "no behavioral change" claim with four checks:

| Check | What it proves | Failure means |
|-------|----------------|---------------|
| **A. Boundary integrity** | No `@chakra-ui/*` import exists outside `components/chakra/` + `components/icons.js` | The library boundary leaked — the Tailwind migration would again need app-wide edits |
| **B. Barrel resolves** | Every Chakra name the barrel re-exports actually exists in the installed `@chakra-ui/react` | A name resolves to `undefined` → `Element type is invalid` at runtime, **with no build error** (the exact trap the barrel header warns about) |
| **C. Consumer contract** | Every name imported `from 'components'` is exported by the barrel | A consumer imports a name the barrel forgot → build break / undefined |
| **D. Move-only diff** | Every git-renamed file changed **only** import/comment lines vs `main` | A "move" secretly changed logic — the one thing a structural PR must never do |

```bash
cd ui
node scripts/verify-migration.js   # exit 0 = clean, 1 = regression
```

Expected on this branch: **6/6 checks pass** (93 Chakra names resolve, 88 distinct consumed names all exported, 177 renamed files all import-only). Any FAIL prints the exact file and symbol.

> This harness is also the permanent guard for the follow-up Tailwind PRs: each step that swaps a primitive inside the library must keep **B** and **C** green, which proves the consumer surface didn't shift.

### Gate 1 — Boundary lint (the guard can't erode)

```bash
cd ui
npx eslint "src/**/*.js"          # must report the no-restricted-imports rule as active with 0 violations
# Negative test — confirm the guard actually bites:
#   temporarily add  import { Box } from '@chakra-ui/react';  to any page,
#   re-run eslint → it MUST error, then revert.
```

### Gate 2 — Production build (integration; **Node 16**)

The real test that webpack resolves every rewritten import and every barrel name. CRA promotes warnings to errors under CI, so build the way CI does:

```bash
cd ui
CI=false yarn build               # matches the deploy workflow; requires ui/.env to exist
```

> ⚠️ This repo targets **Node 16**. On newer Node the CRA build may fail for unrelated reasons — build on Node 16 (or let CI do it) before trusting the result.
>
> Known pre-existing failure (NOT this PR): `yarn test` fails on `CurrentSermon.js`'s `luxon/src/impl/formats` import — fails identically on `main`.

### Gate 3 — Runtime smoke (catches what static checks can't)

Static checks can't see an app-owned component whose default export moved, or a component that renders but throws. Load the highest-traffic routes and watch the console. Compare against the **same routes on `main`** — the bar is *no new* errors, not zero errors.

Routes to hit (cover the moved surface — nav/footer on every page, plus the relocated feature folders):

- `/` (home, NavBar, Footer, MobileNavBar)
- `/about-us` (about/* — StaffSection, TransformationPathwaysSection rename)
- `/sermons` and one `/sermons/:id` (sermons/*, SermonSeries, Pagination)
- `/sermon-notes` (TipTap editor — `TiptapEditor`, `useTextContext`)
- `/give` (giving/*)
- `/profile` (userProfile/* — auth-gated, log in first)
- one admin route, e.g. `/admin/...` (SidebarWithHeader shell + AG Grid)

For each route, confirm in DevTools console:
- **No** `Element type is invalid: expected a string … but got: undefined` (the undefined-export symptom)
- **No** `X is not defined` / `is not a function`
- No *new* network 404s vs `main`

### Gate 4 — Visual parity (the "no visual changes" claim)

For the routes above, take a screenshot on this branch and on `main` at the same viewport (do a desktop **and** a mobile width — this app is responsive-prop heavy) and eyeball-diff. Spot-check the TipTap editor toolbar and an AG Grid table, since those are the heaviest moved components.

### Sign-off checklist

- [ ] Gate 0: `node scripts/verify-migration.js` → 6/6 pass
- [ ] Gate 1: eslint clean + negative test confirms the boundary guard errors
- [ ] Gate 2: `CI=false yarn build` succeeds on Node 16
- [ ] Gate 3: no *new* console errors on the 7 route groups (desktop + mobile)
- [ ] Gate 4: screenshots match `main` on the sampled routes


ksuhartono97 (approved):
Please resolve the merge conflicts haha

PR #1413, opened 2026-07-19
GH-1411: Enable embed support

ksuhartono97:


Midnite1212 (approved):
lgtm, but should confirm on the global web iframe policy. Should be the `X-Frame-Options` or `Content-Security-Policy`

PR #1421, opened 2026-07-28
GH-1404: Migrate UI from CRA to Vite 7 (#1407)

Midnite1212:
* GH-1404: migrate ui from CRA to Vite 7

- react-scripts removed; vite 7 + @vitejs/plugin-react 4 + vitest 3
- vite.config.js: esbuild JSX loader for .js files (no mass rename), /api proxy (replaces setupProxy.js), envPrefix REACT_APP_, PUBLIC_URL defined to '', outDir build/ (CI layout unchanged)
- js-fast-refresh-hint plugin restores Fast Refresh on JSX-in-.js files
- index.html moved to ui/ root, %PUBLIC_URL% refs now root-relative
- 7 process.env.REACT_APP_* sites -> import.meta.env (env files/secrets unchanged)
- eslint + eslint-config-react-app installed explicitly + lint script (CRA bundled them via react-scripts); 0 errors, pre-existing warnings remain
- App.test.js: CRA boilerplate skipped with reason; Vitest wired (jsdom, matchMedia shim); luxon ESM now resolves (old Jest failure gone)
- sass upgraded for Vite's modern Sass API; dead uuid MAX import removed
- CI: Node 16 -> 20; ui/README.md rewritten for Vite (Node 20+ note)

* GH-1404: switch entry to React 18 createRoot

Replaces legacy ReactDOM.render (kept app in React-17 compat mode and logged a console warning). Separate commit so it can be reverted independently of the Vite tooling swap.

* GH-1404: address Copilot review feedback

- App.test.js: skipped test no longer asserts the never-existent 'learn react' text; body is now an honest no-throw smoke placeholder (lint-clean)
- CI: pin node-version to 20.19.0 — Vite 7's engine floor; bare '20' can resolve an older 20.x from the runner toolcache

* GH-1404: address opencode review feedback

- vite.config.js: drop commented-out production-proxy block; instructions moved to ui/README.md
- package.json: remove eslintConfig 'react-app/jest' (runner is Vitest); App.test.js now imports test/expect explicitly from 'vitest'
- index.js: null-guard the GSI script removals in onScriptLoaded (removeChild(null) threw when the gsi/select tag was absent)
- setupTests.js: comment updated for Vitest context
- CI: actions/checkout and setup-node v3 -> v4; fix indentation on the ui install step

Not addressed here by design: axios major bump and pre-existing app bugs (textShadow paren, Signup name/password + errors.country, fontweight, type attrs, duplicate resets) — behavior changes out of scope for the tooling swap; tracked in a follow-up issue.

* GH-1404: fix pages broken by luxon ESM static-method misuse

11 call sites used 'new DateTime.fromISO(...)' / 'new DateTime.now()'. Class static methods are non-constructible, but CRA's production Babel pass transpiled luxon to ES5 functions, so the bogus 'new' silently worked. Vite serves luxon's native-class ESM build, so these threw 'DateTime.now is not a constructor' at runtime — crashing the admin announcements page and silently emptying every announcement-driven list (events page, home events section, sermons, shine involve) via the catch-and-log convention.

Removing 'new' restores the exact CRA-era behavior (the static call's return value was what those expressions evaluated to anyway).

Verified: filter logic replayed against live /api/announcement/get data returns events again; grep confirms zero remaining 'new DateTime.' or similar static-call-with-new patterns; build + lint green.

* GH-1404: harden data handling so malformed records can't white-screen pages

- AnnouncementEditorModal: normalize eventType before feeding CUIAutoComplete. 62 of 73 live announcement records have missing/null eventType (old data shapes) — opening Edit on any of them threw inside CUIAutoComplete and took down the whole admin announcements page. toEventTypeItems() maps legacy string arrays to { label, value } and returns [] for anything non-array; reset now uses [] instead of undefined. Save semantics unchanged — the raw record value is still what gets submitted unless the user edits the field.
- New ErrorBoundary (helpers/ErrorBoundary.js) wrapping the route Switch in MainContainer: any future render crash is contained to the routed page with a friendly fallback + reload, instead of white-screening the app. Keyed by pathname so navigating away resets it.

Verified: replayed the modal's render+submit paths against all 73 live records — 0 crashes (old path crashed on 62); build, lint, tests green.

* GH-1404: address remaining Copilot review feedback

- CI: add explicit "yarn lint" step (vite build, unlike react-scripts, does not run ESLint) so lint errors fail the job before the artifact
- AnnouncementEditorModal: backfill label from value in toEventTypeItems so object records missing a label can't crash CUIAutoComplete

* GH-1404: harden Fast Refresh hint and GSI script cleanup

- vite.config: strip query suffix before matching id, dedupe the hint, return null explicitly when no transform applies
- index.js: match GSI scripts by src prefix (covers ?hl=en copies) and use el.remove() so repeat invocations can't leave duplicates or throw

* GH-1404: fix merge conflicts


ksuhartono97 (commented):
Build succeeded and code looks minimal and targeted for replacing CRA to Vite. LGTM

ksuhartono97 (approved):

PR #1424, opened 2026-08-07
1423: added api for calling user signed up class tracking

Baeck016:


server/api/controllers/classTrackingData/get-signedup-class.js
```diff
@@ -0,0 +1,107 @@
...
+            'record ids:',
+            duplicateIds.join(', '),
+            'using first record:',
+            classDataFetchList[0].id
+          );
+        }
+
+        results.push({
+          formName: form.formName,
+          classTrackingData: classDataFetchList[0],
```
JoshNg12321, line 97: Shouldn't use this as just choosing first record wins -> Think a better way is to check createdAt for the classTrackingData. But imo duplicates for form sign up in the same season shouldn't happen

server/api/controllers/classTrackingData/get-signedup-class.js
```diff
@@ -0,0 +1,107 @@
...
+
+        // There should be one ClassTrackingData record per submission; find the
+        // record(s) for this user and form. If none exists, skip.
+        const classDataFetchList = await ClassTrackingData.find({
+          formId: form.id,
+          userId,
+        });
+        if (!classDataFetchList || classDataFetchList.length === 0) continue;
+
+        if (classDataFetchList.length > 1) {
```
JoshNg12321, line 81: Ideally this shouldn't happen at all for the same season, so have to check whether this an issue

ui/src/pages/forms/FormEditorContainer.js
```diff
@@ -609,6 +618,21 @@ const FormEditorContainer = (props) => {
...
+                                    fontWeight="normal"
+                                    mb="1"
+                                  >
+                                    Class Ending Time
+                                  </FormLabel>
+                                  <Input
+                                    id="classEndingTime"
+                                    type="datetime-local"
+                                    {...register('classEndingTime')}
+                                  />
```
JoshNg12321, line 634: Might need to explain what's this to the stakeholders, maybe ambiguous

server/api/controllers/classTrackingData/get-signedup-class.js
```diff
@@ -0,0 +1,126 @@
...
+        const availableFrom = DateTime.fromISO(form.formAvailableFrom || '');
+        const classEndingTime = DateTime.fromISO(
+          form.classTrackingTemplate?.classEndingTime || ''
+        );
+
+        if (!isCurrentSeason(now, availableFrom, classEndingTime)) continue;
+
+        // A user can have ClassTrackingData from older seasons of the
+        // same reused form, so fetch all and narrow to this season
+        const classDataFetchList = await ClassTrackingData.find({
```
JoshNg12321: Not sure whether there is MongoDB apis that allow you to filter via current season too, and sort by created date so you dont need to add extra code to do it

ui/src/pages/forms/FormEditorContainer.js
```diff
@@ -181,13 +183,20 @@ const FormEditorContainer = (props) => {
       setValue('customEmailSubject', data.customEmailSubject);
       setValue('formAvailableFrom', data.formAvailableFrom);
       setValue('formAvailableUntil', data.formAvailableUntil);
+      setValue(
```
paretypus: can we do some validation before setting the value for class time
e.g.

https://github.com/hmcc-global/hmcchk-web/blob/5c1546555fb160fa6c7b47dc2f31df0df533db79/ui/src/pages/forms/FormEditorContainer.js#L281

server/api/models/Form.js
```diff
@@ -84,11 +84,11 @@ module.exports = {
     },
     classTrackingTemplate: {
       type: 'json',
-      defaultsTo: { courses: [] },
+      defaultsTo: { courses: [], classEndingTime: '' },
```
paretypus: can we add both start and end date time?

server/api/controllers/classTrackingData/get-signedup-class.js
```diff
@@ -0,0 +1,117 @@
...
+      description: 'Something is wrong with your request. Please check it',
+    },
+  },
+
+  fn: async function (inputs, exits) {
+    try {
+      const userId = this.req.user.id;
+
+      // 1. Find submissions for the user to know which forms they've signed up for
+      const submissions = await Submission.find({
```
paretypus: i feel like we don't need to even look at Submissions here right

i think this implementation could be a lot simpler. feel like we can just do

1. get all classTrackingData for a given userId
2. get all the forms related to the above
3. filter formIds from above where class still ongoing (where now >= classStart && now <= classEnd)
4. return classTrackingData where the formId matches above

paretypus (changes requested):

server/api/models/Form.js
Baeck016: Do we actually need a separate classStartingTime? Shouldn't the class start align with the form's available date?

There's a scenario where the class itself hasn't started yet, but there's pre-class material the user needs to complete first, which raises the question: what should classStartingTime actually represent?

paretypus: yes

paretypus:
latest opencode review is fine. the HIgh Prio ones are because the implementation sits at a different PR

paretypus (approved):

PR #1437, opened 2026-08-23
GH-1436: fill live sermon passage from published sermon notes

### LINKED ISSUE #1436: Auto-fill /online sermon passage from the published sermon notes
The Sermon Passage field in the Live Sermon Manager is typed by hand each week even though the same passage is already recorded on that Sunday's sermon notes. It should be filled in from the sermon notes for the matching service date, taking the published record when more than one exists for that date.

Midnite1212:
Closes #1436

The Sermon Passage field in the Live Sermon Manager is now filled in from the sermon notes matching the selected service date, taking the published record with the newest `sermonId` when several exist for that date, and staying editable as an override.

### DIFF
--- a/server/api/controllers/sermonNotesParent/get-sermon-notes-parent.js
+++ b/server/api/controllers/sermonNotesParent/get-sermon-notes-parent.js
@@ -8,6 +8,15 @@ module.exports = {
       required: false,
       type: 'string',
     },
+    date: {
+      required: false,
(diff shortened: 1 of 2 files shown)


ksuhartono97 (changes requested):
Code seems to do what it needs to. Please do functional testing + address the comments from copilot about file casing.

